PR #10, opened 2016-09-09
Pass compiler flags down to the Tox test environment

elventear:
In order to test a module that contains C/C++ code, it might be necessary to tweak the compilation parameters according to the environment where tox is running. Currently it does not pass compilation flags like `CFLAGS` and the like.

A way to override this manually is:

```
> export TOX_TESTENV_PASSENV='CFLAGS CPPFLAGS LDFLAGS'
> python setup.py tox
```

But I am afraid this hidden and requires people to search for the information. I think we should do the right thing and do not tamper with those environment variables if they have been set for a reason.

### DIFF
--- a/pykern/pksetup.py
+++ b/pykern/pksetup.py
@@ -260,7 +260,7 @@ class Tox(setuptools.Command, object):
 [tox]
 envlist={pyenv}
 [testenv]
-passenv=PKSETUP_PKDEPLOY_IS_DEV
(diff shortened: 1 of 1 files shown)


elventear:
Creating this a PR, to get some feedback on the change, since I am not that familiar with all the machinery relating to distutils, tox, etc.

robnagler:
I'm a little confused about the need for the change. Is this necessary for
rslinac?

elventear:
Yes. I need it build it when using boost installed in standard path. There is no way for me to influence the compiler flags.

robnagler:
Then sounds great. :)

PR #45, opened 2019-11-15
Atomic write

e-carlin:


pykern/pkio.py
```diff
@@ -18,11 +18,26 @@
 import re
 import shutil
 import six
+import tempfile
 
 #: used during unit testing see ``pykern.pkunit.save_chdir``
 pkunit_prefix = None
 
 
+def atomic_write(filename, content, mode='w'):
```
robnagler: No need for mode.

pykern/pkio.py
```diff
@@ -18,11 +18,26 @@
...
+    """Writes to file atomically
+
+    Args:
+        filename (str or py.path.Local): File to write to
+    """
+    with tempfile.NamedTemporaryFile(
```
robnagler: I think we should just name the file with a suffix, e.g. '.pkio-tmp' and not use a library.

pykern/pkio.py
```diff
@@ -18,11 +18,26 @@
...
 import shutil
 import six
+import tempfile
 
 #: used during unit testing see ``pykern.pkunit.save_chdir``
 pkunit_prefix = None
 
 
+def atomic_write(filename, content, mode='w'):
+    """Writes to file atomically
```
robnagler: Need text that is distinct from the name, e.g. Overwrites an existing file via rename to ensure integrity.

pykern/pkio.py
robnagler: I think we should switch to "path" to indicate it might be a py.path. You can change write_text, too.

Make contents and content consistent, too.

pykern/pkio.py
```diff
@@ -18,11 +18,26 @@
...
+
+    Args:
+        filename (str or py.path.Local): File to write to
+    """
+    with tempfile.NamedTemporaryFile(
+        mode=mode, dir=os.path.dirname(str(filename))) as f:
```
robnagler: Once converted to a py.path, you can then change the suffix to pkio-tmp, and it'll be good enough, the new code would then look like:

```py
n = p.new(ext='pkio-tmp')
n.write(contents)
n.rename(p)
```

pykern/pkio.py
```diff
@@ -18,11 +18,26 @@
...
+        f.write(content)
+        if os.path.exists(str(filename)):
+            os.remove(str(filename))
```
robnagler: This operation defeats the purpose of the function. `rename` is atomic (even by py.path). This isn't atomic, because the remove happens then the link. 

Some people may say that remove and link are necessary, because the file may not be writable, but the directory might be writable. That's correct, but not the purpose of this particular call which is not to "write no matter what", but to "write atomically". This is in a safe environment from the perspective of "no matter what", but not safe from process exits or system crashes or other types of asynchronous events.

e-carlin: Doh, of course.

PR #91, opened 2021-02-12
create abstraction for module_functions

e-carlin:


pykern/pkinspect.py
```diff
@@ -116,7 +116,7 @@ def caller(ignore_modules=None, exclude_first=True):
             del frame
 
 
-def caller_module():
+def caller_module(exclude_first=True):
```
robnagler, line 119: Description needs to be changed to describe exclude_first.

pykern/pkinspect.py
```diff
@@ -126,10 +126,13 @@ def caller_module():
 
     Will raise exception if calling from __main__
 
+    Args:
+        exclude_first (bool): skip first module found [True]
```
robnagler, line 130: calling module, not the caller's caller.

pykern/pkinspect.py
```diff
@@ -191,6 +194,23 @@ def module_name_split(obj):
     return n.split('.');
 
 
+def module_functions(func_prefix, module=None):
+    """Get all module level functions with a prefix
```
robnagler, line 198: "Get all functions in `module` starting with `func_prefix`"

pykern/pkinspect.py
```diff
@@ -191,6 +194,23 @@ def module_name_split(obj):
...
+
+    Args:
+        func_prefix (str): the prefix of function names to get
+        module (object): a module to get functions from (caller module if None)
```
robnagler, line 202: "calling module"

robnagler (changes requested):

pykern/pkinspect.py
robnagler: Needs a test

pykern/pkinspect.py
e-carlin: b29cdf7

pykern/pkinspect.py
e-carlin: b29cdf7

pykern/pkinspect.py
e-carlin: I'm not sure what your saying?

pykern/pkinspect.py
e-carlin: b29cdf7

pykern/pkinspect.py
robnagler: The documentation is unclear. exclude_first means to not do what the documentation says it is supposed to do so the documentation should read something like:

```txt
    """Returns `Call` for caller's caller with certain exclusions

    Will not return the same module as the calling module, that is,
    will iterate until a new module is found, when `exclude_first` is
    truthy. If `ignore_modules` is defined, will skip over all call frames
    for those modules as well.

    Note: may return ``__main__`` module.

    Will raise exception if calling from ``__main__``

    Args:
        ignore_modules (list): other modules (objects) to exclude [None]
        exclude_first (bool): skip over calls from calling module [True]

    Returns:
        Call: keys: filename, lineno, name, module
```

robnagler: fix later, because both functions need updating, and better to get this out the door.

robnagler (approved):

PR #96, opened 2021-03-03
Fix #93: Only backup issues updated in last 7 days

### LINKED ISSUE #93: can github.bkp cache issues?
Need to detect if "updated" is always updated when an issue changes.

e-carlin:


pykern/pkcli/github.py
```diff
@@ -377,8 +377,11 @@ def _issue(i):
                 if p:
                     j['review_comments'] = [_trim_body(c) for c in p.review_comments()]
                 pkjson.dump_pretty(j, filename=d.join(str(i.number) + '.json'))
+            for i in _try(lambda: list(repo.issues(
```
robnagler, line 380: Did you test if this actually is correct, e.g. if you edit an issue, delete a comment, etc. All those things must update "since".

Also, we need to backup everything the first time, and then keep a repo so it has to be more like the [code in clone](https://github.com/radiasoft/pykern/blob/d43b9f20b8829a5a6e076dd6032dffcbe69cee50/pykern/pkcli/github.py#L346)

e-carlin: > Did you test if this actually is correct, e.g. if you edit an issue, delete a comment, etc. All those things must update "since".

They all seem to update since. The only one that I tested that doesn't is adding an emoji to a comment.

robnagler: Ugh, I know that seems trivial, but probably +1's are not being counted. That's not huge, but it's annoying. Just to be clear, updating any comment (not just the first one, which is special) causes the "since" to work?

e-carlin: > Ugh, I know that seems trivial, but probably +1's are not being counted. That's not huge, but it's annoying. Just to be clear, updating any comment (not just the first one, which is special) causes the "since" to work?

Yes, updating a comment that isn't the first works.

https://github.com/radiasoft/pykern/pull/96/commits/ac23059c0d22caa7b577da6b7a73b4ea15e1e9b9 for getting all on the first go and then after that pulling in the prev backup and merging with new stuff.

robnagler (approved):

PR #104, opened 2021-08-06
Fix radiasoft/sirepo#3732: Update pkresource to search additional packages

e-carlin:
Sirepo needs to be able to load resources that live outside of its namespace.
This updates pkresource to support searching for resources beyond the
caller_context / caller_module.

The additional packages are searched first so resources found in them can
override the resources in Sirepo (or any caller context/module).

### DIFF
--- a/pykern/pkresource.py
+++ b/pykern/pkresource.py
@@ -10,10 +10,8 @@ from __future__ import absolute_import, division, print_function
 import errno
 import glob as builtin_glob
 import importlib
-import inspect
 import os.path
 import pkg_resources
-import re
 
 # TODO(e-carlin): discuss with rn if ok to import pkio
 from pykern import pkinspect
--- a/pykern/pkresource.py
+++ b/pykern/pkresource.py
@@ -31,8 +31,6 @@ def filename(relative_filename, caller_context=None, packages=None, relpath=Fals
     Returns:
         str: absolute or relative (to package_data) path of the resource file
     """
-    if caller_context and packages:
-        raise ValueError(f'Use only one of caller_context={caller_context} and packages={packages}.')
     assert not os.path.isabs(relative_filename), \
         'must not be an absolute file name={}'.format(relative_filename)
     a = []
@@ -44,10 +42,7 @@ def filename(relative_filename, caller_context=None, packages=None, relpath=Fals
                     os.path.join(pkg_resources.resource_filename(p, ''), pksetup.PACKAGE_DATA),
                 ).bestrelpath(pkio.py_path(f)))
             return f
-    msg = f'unable to locate in packages={a}'
-    if '__main__' in a:
-        msg += '; do not call module as a program'
-    raise IOError(errno.ENOENT, msg, relative_filename)
+    _raise_no_file_found(a, relative_filename)
 
 
 def glob_files(relative_path, caller_context=None, packages=None):
@@ -63,12 +58,18 @@ def glob_files(relative_path, caller_context=None, packages=None):
     if caller_context and packages:
         raise ValueError(f'Use only one of caller_context={caller_context} and packages={packages}.')
     res = []
-    for f, _ in _files(relative_path, caller_context, packages):
+    a = []
+    for f, p in _files(relative_path, caller_context, packages):
+        a.append(p)
         res.extend(glob.glob(f))
+    if not res:
+        _raise_no_file_found(a, relative_path)
     return res
 
 
 def _files(path, caller_context, packages):
+    if caller_context and packages:
+        raise ValueError(f'Use only one of caller_context={caller_context} and packages={packages}.')
     for p in list(map(
         lambda m: pkinspect.root_package(importlib.import_module(m)),
         packages or \
@@ -84,3 +85,10 @@ def _files(path, caller_context, packages):
             p,
             os.path.join(pksetup.PACKAGE_DATA, path),
         ), p
+
+
+def _raise_no_file_found(packages, path):
+    msg = f'unable to locate in packages={packages}'
+    if '__main__' in packages:
+        msg += '; do not call module as a program'
+    raise IOError(errno.ENOENT, msg, path)
--- a/pykern/pkresource.py
+++ b/pykern/pkresource.py
@@ -42,10 +42,10 @@ def filename(relative_filename, caller_context=None, additional_packages=None, r
                     os.path.join(pkg_resources.resource_filename(p, ''), pksetup.PACKAGE_DATA),
                 ).bestrelpath(pkio.py_path(f)))
             return f
-    msg = f'unable to locate resource={relative_filename} in packages={a}'
+    msg = f'unable to locate in packages={a}'
     if '__main__' in a:
         msg += '; do not call module as a program'
-    raise IOError(errno.ENOENT, msg)
+    raise IOError(errno.ENOENT, msg, relative_filename)
 
 
 def glob(relative_path, caller_context=None, additional_packages=None):
--- a/pykern/pkresource.py
+++ b/pykern/pkresource.py
@@ -71,7 +71,8 @@ def glob_files(relative_path, caller_context=None, packages=None):
 def _files(path, caller_context, packages):
     for p in list(map(
         lambda m: pkinspect.root_package(importlib.import_module(m)),
-        packages or pkinspect.root_package(caller_context if caller_context else pkinspect.caller_module()),
+        packages or \
+            [pkinspect.root_package(caller_context if caller_context else pkinspect.caller_module())],
     )):
         # TODO(e-carlin): using pkg_resources is discouraged
         # https://setuptools.readthedocs.io/en/latest/pkg_resources.html
--- a/pykern/pkresource.py
+++ b/pykern/pkresource.py
@@ -8,7 +8,7 @@ from __future__ import absolute_import, division, print_function
 
 # Root module: Import only builtin packages so avoid dependency issues
 import errno
-import glob as builtin_glob
+import glob
 import importlib
 import os.path
 import pkg_resources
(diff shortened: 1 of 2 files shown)


pykern/pkresource.py
```diff
@@ -8,34 +8,82 @@
...
+import glob as builtin_glob
+import importlib
 import inspect
 import os.path
 import pkg_resources
 import re
 
+# TODO(e-carlin): discuss with rn if ok to import pkio
 from pykern import pkinspect
+from pykern import pkio
```
e-carlin, line 20: @robnagler  is this ok?

pykern/pkresource.py
```diff
@@ -8,34 +8,82 @@
...
+        if os.path.exists(f):
+            if relpath:
+                f = str(pkio.py_path(
+                    os.path.join(pkg_resources.resource_filename(p, ''), pksetup.PACKAGE_DATA),
+                ).bestrelpath(pkio.py_path(f)))
+            return f
+    msg = f'unable to locate resource={relative_filename} in packages={a}'
+    if '__main__' in a:
+        msg += '; do not call module as a program'
+    raise IOError(errno.ENOENT, msg)
```
e-carlin, line 50: @robnagler I suppose this is a non backwards compat change. The structure of the message is different and no filename. Is that ok?

pykern/pkresource.py
```diff
@@ -8,34 +8,82 @@
 
 # Root module: Import only builtin packages so avoid dependency issues
 import errno
+import glob as builtin_glob
```
e-carlin, line 11: @robnagler thoughts on using as?

pykern/pkresource.py
robnagler: This is fine, because it is an error message, and if we don't have a file name, we don't...

pykern/pkresource.py
robnagler: This is a smell. Rename your method (see https://github.com/radiasoft/sirepo/pull/3747#discussion_r685268361) to sorted_glob or something that doesn't collide.

pykern/pkresource.py
robnagler: Yes, but add a comment about limiting imports in pkio.

pykern/pkresource.py
```diff
@@ -8,34 +8,82 @@
...
+    # files of the same name
+    for p in list(map(
+        lambda m: pkinspect.root_package(importlib.import_module(m)),
+        additional_packages or [],
+    )) + [
+        pkinspect.root_package(
+            caller_context if caller_context else pkinspect.caller_module(),
+        )
+    ]:
+        # TODO(e-carlin): using pkg_resources is discouraged
```
robnagler, line 80: When they say discouraged, it probably isn't it...

pykern/pkresource.py
```diff
@@ -8,34 +8,82 @@
...
+    for f, _ in _files(relative_path, caller_context, additional_packages):
+        res.extend(builtin_glob.glob(f))
     return res
+
+
+def _files(path, caller_context, additional_packages):
+    # Search additional_packages first so they can possibly override caller context/module
+    # files of the same name
+    for p in list(map(
+        lambda m: pkinspect.root_package(importlib.import_module(m)),
```
robnagler, line 73: It will be simpler if this is `search_path`.

robnagler (changes requested):

pykern/pkresource.py
e-carlin: 3260b72f49

pykern/pkresource.py
e-carlin: f5f61ee

pykern/pkresource.py
e-carlin: ccca032

pykern/pkresource.py
e-carlin: ccca032

pykern/pkresource.py
e-carlin: :/

e-carlin:
@robnagler ready for another review.

pykern/pkresource.py
```diff
@@ -8,34 +8,78 @@
...
+    """Find all paths that match the relative path
+
+    Args:
+        relative_path(str): Path relative to package_data directory.
+        caller_context (object): Any object from which to get the `root_package`.
+        packages (List[str]): Packages to search.
+    Returns:
+        [str]: absolute paths of the matched files
+    """
+    if caller_context and packages:
```
robnagler, line 63: This is the same code as `filename()` so share it. If it comes up empty, might be for two reasons: no packages or really not found. glob probably should throw an exception in that case. Shouldn't be called from `__main__` either.

pykern/pkresource.py
```diff
@@ -8,34 +8,78 @@
...
+        packages (List[str]): Packages to search.
+        relpath (bool): If True path is relative to package package_data dir
 
     Returns:
-        str: absolute path of the resource file
+        str: absolute or relative (to package_data) path of the resource file
     """
-    pkg = pkinspect.root_package(
-        caller_context if caller_context else pkinspect.caller_module())
+    if caller_context and packages:
```
robnagler, line 34: Move the assertion in `_files()`.

robnagler (changes requested):

pykern/pkresource.py
e-carlin: a60907c

pykern/pkresource.py
e-carlin: a60907c

e-carlin:
@robnagler ready for another review.

pykern/pkresource.py
```diff
@@ -8,34 +8,87 @@
...
+    """Find all paths that match the relative path
+
+    Args:
+        relative_path(str): Path relative to package_data directory.
+        caller_context (object): Any object from which to get the `root_package`.
+        packages (List[str]): Packages to search.
+    Returns:
+        [str]: absolute paths of the matched files
+    """
+    if caller_context and packages:
```
robnagler, line 58: remove

pykern/pkresource.py
```diff
@@ -8,34 +8,87 @@
...
+        packages or \
+            [pkinspect.root_package(caller_context if caller_context else pkinspect.caller_module())],
+    )):
+        # TODO(e-carlin): using pkg_resources is discouraged
+        # https://setuptools.readthedocs.io/en/latest/pkg_resources.html
+        # But, as of py3.7 importlib.resources doesn't offer a good API
+        # for accessing directories
+        # https://docs.python.org/3/library/importlib.html#module-importlib.resources
+        # https://gitlab.com/python-devs/importlib_resources/-/issues/58
+        yield pkg_resources.resource_filename(
```
robnagler, line 84: formatting will be simpler with extra parens:

```py
        yield (
            pkg_resources.resource_filename(
                p,
                os.path.join(pksetup.PACKAGE_DATA, path),
            ), 
            p,
        )
```

pykern/pkresource.py
```diff
@@ -8,34 +8,87 @@
...
+        a.append(p)
+        res.extend(glob.glob(f))
+    if not res:
+        _raise_no_file_found(a, relative_path)
     return res
+
+
+def _files(path, caller_context, packages):
+    if caller_context and packages:
+        raise ValueError(f'Use only one of caller_context={caller_context} and packages={packages}.')
```
robnagler, line 72: remove `.` at the end of the string. We don't use them typically, and in this case, it would look strange.

pykern/pkresource.py
```diff
@@ -8,34 +8,87 @@
...
-    res = pkg_resources.resource_filename(pkg, fn)
-    if not os.path.exists(res):
-        msg = 'resource does not exist for pkg=' + pkg
-        if pkg == '__main__':
-            msg += '; do not call module as a program'
-        raise IOError((errno.ENOENT, msg, res))
+    a = []
+    for f, p in _files(relative_filename, caller_context, packages):
+        a.append(p)
+        if os.path.exists(f):
```
robnagler, line 39: To avoid extra indent, use if not:

```py
if not os.path.exists(f):
    continue
```

pykern/pkresource.py
```diff
@@ -8,34 +8,87 @@
...
+            if relpath:
+                f = str(pkio.py_path(
+                    os.path.join(pkg_resources.resource_filename(p, ''), pksetup.PACKAGE_DATA),
+                ).bestrelpath(pkio.py_path(f)))
+            return f
+    _raise_no_file_found(a, relative_filename)
+
+
+def glob_files(relative_path, caller_context=None, packages=None):
+    """Find all paths that match the relative path
```
robnagler, line 49: Find all paths that match the relative path __in all packages__

robnagler (changes requested):

pykern/pkresource.py
e-carlin: 0888512

pykern/pkresource.py
e-carlin: 0888512

pykern/pkresource.py
e-carlin: 0888512

pykern/pkresource.py
e-carlin: 0888512

pykern/pkresource.py
e-carlin: 0888512

e-carlin:
@robnagler ready for another review.

pykern/pkresource.py
e-carlin: Had to change this to not raise on not found 685c96e

pykern/pkresource.py
```diff
@@ -8,34 +8,97 @@
...
-import re
 
+# TODO(e-carlin): discuss with rn if ok to import pkio
 from pykern import pkinspect
+from pykern import pkio
 from pykern import pksetup
 
 
-def filename(relative_filename, caller_context=None):
+def file_path(relative_filename, caller_context=None, packages=None):
```
robnagler, line 22: This will break things. Leave old routine until everything is fixed. I was just using filename() in rsdockerspawner, and don't want to coordinate.

robnagler (changes requested):

robnagler (approved):

PR #107, opened 2021-09-23
Fix #106 - pkre apply pkcompat to actual

### LINKED ISSUE #106: pkre should implement its doc:  actual (object): run-time value
`actual` should be converted to a string. pkcompat should be used for bytes.

mkeilman:
Tested against 'string', b'string', and u'string'

### DIFF
--- a/pykern/pkunit.py
+++ b/pykern/pkunit.py
@@ -351,7 +351,8 @@ def pkre(expect_re, actual, flags=re.IGNORECASE + re.DOTALL):
         actual (object): run-time value
         flags: passed on to re.search [IGNORECASE + DOTALL]
     """
-    if not re.search(expect_re, actual, flags=flags):
+    from pykern import pkcompat
+    if not re.search(expect_re, pkcompat.from_bytes(actual), flags=flags):
         pkfail('expect_re={} != actual={}', expect_re, actual)
 
 
--- a/pykern/pkunit.py
+++ b/pykern/pkunit.py
@@ -6,6 +6,7 @@ u"""Useful operations for unit tests
 """
 from __future__ import absolute_import, division, print_function
 from pykern import pkcollections
+from pykern import pkcompat
 from pykern import pkinspect
 from pykern import pkio
 # defer importing pkconfig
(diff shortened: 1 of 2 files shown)


pykern/pkunit.py
```diff
@@ -351,7 +351,8 @@ def pkre(expect_re, actual, flags=re.IGNORECASE + re.DOTALL):
         actual (object): run-time value
         flags: passed on to re.search [IGNORECASE + DOTALL]
     """
-    if not re.search(expect_re, actual, flags=flags):
+    from pykern import pkcompat
```
robnagler, line 354: This import can be at the top and should be.

pykern/pkunit.py
```diff
@@ -351,7 +351,8 @@ def pkre(expect_re, actual, flags=re.IGNORECASE + re.DOTALL):
...
+    if not re.search(expect_re, pkcompat.from_bytes(str(actual)), flags=flags):
```
robnagler, line 355: I don't think this is correct `pkcompat.from_bytes(str(actual))`. from_bytes takes a bytes argument but str returns a string. You can remove the str call, and I think it should give you what you want. pkcompat.from_bytes is robust to the fact that it might be given a string.

robnagler (changes requested):
> Tested against 'string', b'string', and u'string'

Please write a test case. There is already a test for pkunit. Verify the test fails without your change, and then add your change.

pykern/pkunit.py
mkeilman: 7918c4a

pykern/pkunit.py
mkeilman: 7918c4a

Never sure when to place at the top or in a function...

robnagler: It's when there is likely to be a dependency issue. pkcompat is specifically designed to not be a dependency problem. In general, at the top is the right default. In Sirepo, there is more complexity with circular imports, especially with all the dependencies, e.g. h5py.

mkeilman:
> > Tested against 'string', b'string', and u'string'
> 
> Please write a test case. There is already a test for pkunit. Verify the test fails without your change, and then add your change.

cfbcfa9

tests/pkunit_test.py
```diff
@@ -127,6 +127,13 @@ def test_pkok():
         assert 'pkunit_test.py:{}:test_pkok xyzzy 333 abc'.format(lineno) in str(e.args)
 
 
+def test_pkre_convert():
+    from pykern.pkunit import pkre
+    r = r'A'
+    for s in ('A', b'A', r'A', u'A'):
```
robnagler, line 133: As long as I'm being nitpicky.... Only two of these cases are necessary. `r` doesn't actually change the type of the string. It is akin to `f` which changes how the string is interpreted. `u` is a no-op in py3. The `r` variable is unnecessary, and the `r` on the string is unnecessary. In a non-test environment where performance might be an issue, you could compile the regex outside the loop, but that would complicate the code to no effect so better to simplify.

robnagler (approved):

PR #112, opened 2021-11-29
add sorting by key

e-carlin:


pykern/pkio.py
```diff
@@ -225,16 +225,20 @@ def save_chdir(dirname, mkdir=False, is_pkunit_prefix=False):
...
         path (py.path.Local or str): pattern
+        key (str): name of a function of path used to sort
 
     Returns:
         list: py.path.Local objects
     """
-    return sorted(py_path(f) for f in glob.glob(str(path)))
+    return sorted(
+        (py_path(f) for f in glob.glob(str(path))),
+        key=(lambda p: getattr(p, key)()) if key else None,
```
robnagler, line 240: Good. I almost thought it would be better to remove the invariant, but that would be reaching inside the py.path abstraction so better this way. :)

tests/pkio_test.py
```diff
@@ -116,7 +116,7 @@ def test_walk_tree_and_sorted_glob():
             'When walking tree with file_re, should only return matching files'
         assert [expect[0]] == list(pkio.walk_tree('.', '^d1')), \
             'When walking tree with file_re, file to match does not include dir being searched'
-        assert pkio.sorted_glob('*[42]') == [py.path.local(f) for f in ('d2', 'd4')]
+        assert pkio.sorted_glob('*[42]', key='mtime') == [py.path.local(f) for f in ('d4', 'd2')]
```
robnagler, line 119: mtime is not going to be very good for this situation, because all files are created at the same time and Linux only keeps mtime in seconds. This would be better:

```py
pkio.sorted_glob('*/*/*', key='basename') == expect
```

robnagler (changes requested):

tests/pkio_test.py
robnagler: basename is a problem but it's a good test, because it is a useful. It is not a callable so the key code just needs to test whether the object is callable or not.

e-carlin: Good thinking. 66fb4e1

robnagler (approved):

PR #132, opened 2022-02-28
Fix #117 setuptools must be less than 57

### LINKED ISSUE #117: pksetup.pkdeploy fails with SyntaxError in itertools
Caused by https://github.com/pypa/setuptools/issues/2687

The error is:

```txt
raise type(self._exception), self._exception, self._traceback
                                     ^
      SyntaxError: invalid syntax
```

This only happens because we are using tox, which does a clean install to build an sdist. I think we need to dump tox to build sdists. But documenting this here.

robnagler:
see https://github.com/pypa/setuptools/issues/2687

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
         'pytz>=2015.4',
         'ruamel.yaml>=0.16.0',
         'requests>=2.18',
-        'setuptools>=20.3',
+        'setuptools<57',
         'six>=1.9',
         'Sphinx>=1.3.5',
         'twine>=1.9',


e-carlin (approved):
Seems fine. I don't understand what the point of all of the checks for python version are if they don't work https://github.com/pypa/setuptools/blob/19bfd4d13755011a027013a63a637f02bda0c0cd/setup.cfg#L27 and https://github.com/agronholm/pythonfutures/blob/acba785c8a5c331edce05dfa760540b8ea8a81f1/setup.py#L8

robnagler:
This is why you feature test, and don't version test. If the code were properly written, it would isolate old syntax in a separate module that is loaded. If it fails, then the syntax isn't supported. However, that's not what people do.

PR #136, opened 2022-03-14
Fix #135: constants for rslaser and rsbeam

### LINKED ISSUE #135: Add rsbeams and rslaser constants to pkconst
Constants used by rsbeams and rslaser need to be added to pykern see [this issue](https://github.com/radiasoft/rslaser/issues/55)

gurhar1133:
Constants from:
https://github.com/radiasoft/rsbeams/blob/master/rsbeams/rsphysics/rsconst.py
https://github.com/radiasoft/rslaser/blob/master/rslaser/utils/constants.py
added to pkconst.py

See this [issue](https://github.com/radiasoft/rslaser/issues/55)


pykern/pkconst.py
```diff
@@ -31,3 +33,33 @@
 
 #: Copied from numconv, which copied from RFC1924
 BASE62_CHARS = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+TWO_PI = 2 * math.pi
```
robnagler, line 36: Document each of these with `#:` prefixes. See two lines above. It's fine for the comment to be simple:

```py
#: 2 * PI
TWO_PI = 2 * math.pi
```

Sort constants alphabetically except if there's a dependency.

All constants are upper case since that's PEP8 compliant.

I would suggest renaming TWO_PI to PI_2, but I know that people may be particular.

pykern/pkconst.py
```diff
@@ -31,3 +33,33 @@
...
+RT_TWO_PI = math.sqrt(2*math.pi)
+RT_2_OVER_PI = math.sqrt(2/math.pi)
+pi = math.pi
+RT_2 = math.sqrt(2.)
+TWO_PI = 2 * math.pi
+RT_TWO_PI = math.sqrt(2*math.pi)
+RT_2_OVER_PI = math.sqrt(2/math.pi)
+c = scipy.constants.c                  # speed of light [m/s]
+c_SQ = scipy.constants.c**2
```
robnagler, line 45: no mixed case

pykern/pkconst.py
```diff
@@ -31,3 +33,33 @@
...
+epsilon_0 = scipy.constants.epsilon_0  # permittivity of free space
+m_p = scipy.constants.m_p              # proton mass [kg]
+m_e = scipy.constants.m_e              # electron mass [kg]
+e = scipy.constants.e                  # fundamental electric charge [C] (positive)
+MKS_factor = 1./(4.*math.pi*scipy.constants.epsilon_0)
+KG_to_EV = c_SQ / e          # convert mass [kg] to effective energy [eV]
+EV_to_ERG = 1.602e-12        # convert energy from eV to erg
+C_to_STATC = scipy.constants.c * 10.   # convert charge from coulombs to statcoulombs
+e_CGS = scipy.constants.e * C_to_STATC
+c_CGS = scipy.constants.c * 100
```
robnagler, line 57: Missing `.`?

pykern/pkconst.py
```diff
@@ -31,3 +33,33 @@
...
+m_e_CGS = scipy.constants.m_e * 1000.
+m_p_CGS = scipy.constants.m_p * 1000.
+m_e_EV = scipy.constants.m_e * KG_to_EV
+m_p_EV = scipy.constants.m_p * KG_to_EV
+C_SQ = scipy.constants.c**2
+C_INV  = 1./scipy.constants.c
```
robnagler, line 63: reuse the constant C, not scipy.constants.c, so easier to read.

pykern/pkconst.py
```diff
@@ -31,3 +33,33 @@
 
 #: Copied from numconv, which copied from RFC1924
 BASE62_CHARS = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+TWO_PI = 2 * math.pi
+RT_TWO_PI = math.sqrt(2*math.pi)
+RT_2_OVER_PI = math.sqrt(2/math.pi)
+pi = math.pi
+RT_2 = math.sqrt(2.)
+TWO_PI = 2 * math.pi
+RT_TWO_PI = math.sqrt(2*math.pi)
```
robnagler, line 42: Be consistent with spacing. below you have eg `c_SQ / e` (with spaces). Here there are no spaces. The software group normally spaces things like other Python code: `math.sqrt(2 * math.pi)`. In this case, you really want, `math.sqrt(TWO_PI)`, because it's already defined.

pykern/pkconst.py
```diff
@@ -31,3 +33,33 @@
...
+RT_2_OVER_PI = math.sqrt(2/math.pi)
+c = scipy.constants.c                  # speed of light [m/s]
```
robnagler, line 44: We do not use side comments. They do not show up in Sphinx/Napoleon. They are hard to manage, because they are dependent on the font and the width of the strings. If names change, they are misaligned. Our coding style avoids misalignments when names change.

robnagler (changes requested):

pykern/pkconst.py
gurhar1133: What do you mean?

robnagler: Dot after the 100 like the other numeric values have: `C_CGS = C * 100.`

pykern/pkconst.py
```diff
@@ -33,27 +33,51 @@
...
-m_e_CGS = m_e * 1000.
-m_p_CGS = m_p * 1000.
+#: C * 100
+C_CGS = C * 100.
+#: m_e * 1000
+M_E_CGS = m_e * 1000.
+#: m_p * 1000
+M_P_CGS = m_p * 1000.
+#: m_e * C_SQ / e
 m_e_EV = m_e * KG_to_EV
```
gurhar1133, line 79: @robnagler @bruhwiler there seems to be a conflict with not using mixed case and with this line:

https://github.com/radiasoft/pykern/blob/f149c5d23696c3293e650147e5c70be3e1152a7f/pykern/pkconst.py#L79

 and this one:

https://github.com/radiasoft/pykern/blob/f149c5d23696c3293e650147e5c70be3e1152a7f/pykern/pkconst.py#L83

If we dont use mixed case then we have to rename one of these constants.

pykern/pkconst.py
bruhwiler: Changing TWO_PI to PI_2 would require other changes for the sake of being consistent.

More importantly, scientists might wonder what it means and have to double check.
In contrast, TWO_PI will be unambiguous for scientists and mathematicians.

I recommend keeping the name as is.

pykern/pkconst.py
bruhwiler: The two variants are the same, except perhaps for a minus sign.
I recommend keeping M_E_EV = m_e * C_SQ / (-e) and deleting the other.

bruhwiler:
Here, I am contradicting the decision I made when originally writing rsconst.py but it now seems better that users should not go to RadiaSoft constants when they would just be duplicating something available in SciPy.

For example, I suggest that  m_e = scipy.constants.m_e  and all analogous cases should be removed.

robnagler:
> For example, I suggest that m_e = scipy.constants.m_e and all analogous cases should be removed.

I agree.

I still would like all the constants to be consistent uppercase. [PEP8 is clear about this](https://peps.python.org/pep-0008/#constants).

bruhwiler:
I agree that all of the constants defined by RadiaSoft should be all uppercase.

pykern/pkconst.py
```diff
@@ -5,6 +5,8 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
+import math
+import scipy.constants
```
e-carlin, line 9: pykern doesn't list scipy in the install_requires. On my mac I now see this
```python
pykern (135-rsbeams-rslaser-constants)$ python -c 'import pykern.pkdebug'
  File "<string>", line 1, in <module>
  File "/Users/ecarlin/src/radiasoft/pykern/pykern/pkdebug.py", line 71, in <module>
    from pykern import pkconfig
  File "/Users/ecarlin/src/radiasoft/pykern/pykern/pkconfig.py", line 75, in <module>
    from pykern import pkconst
  File "/Users/ecarlin/src/radiasoft/pykern/pykern/pkconst.py", line 9, in <module>
    import scipy.constants
ModuleNotFoundError: No module named 'scipy'
```

What about moving these consts to some other module (ex `pykern/pkmath.py`) and we are careful to not use that in context where scipy may not be installed?

robnagler: That's a good point. Do we want pykern to require scipy? I think so, but I could be convinced differently. That's the "big library" problem we had with BOP. scipy has the big library problem, too, but people don't seem to mind.

e-carlin: > Here, I am contradicting the decision I made when originally writing rsconst.py but it now seems better that users should not go to RadiaSoft constants when they would just be duplicating something available in SciPy.

> For example, I suggest that m_e = scipy.constants.m_e and all analogous cases should be removed.

Will fix this.

gurhar1133:
> Here, I am contradicting the decision I made when originally writing rsconst.py but it now seems better that users should not go to RadiaSoft constants when they would just be duplicating something available in SciPy.
> 
> For example, I suggest that m_e = scipy.constants.m_e and all analogous cases should be removed.

I can remove those cases, but it does not look to me like either of m_e_EV or M_E_EV (from previous comment) fall into that category. If so, there is still a naming clash

bruhwiler:
@gurhar1133 -- I agree with you.

As noted in a different thread on this PR:
"The two variants are the same, except perhaps for a minus sign.
I recommend keeping M_E_EV = m_e * C_SQ / (-e) and deleting the other."

pykern/pkconst.py
bruhwiler: @gurhar1133 is correct.  The problem of dependence on scipy must still be addressed.

e-carlin: I don't think we should but also could be convinced otherwise:). Here are some reasons:
- scipy [depends on py38](https://github.com/scipy/scipy/blob/b5d8bab88af61d61de09641243848df63380a67f/pyproject.toml#L58). We run py37. We want to upgrade but who knows when that will happen. I don't want to be forced to upgrade for a dep that provides as little to pykern as scipy does. But maybe being forced would be good...
- scipy [depends on numpy](https://github.com/scipy/scipy/blob/b5d8bab88af61d61de09641243848df63380a67f/pyproject.toml#L25). What version exactly seems confusing. Again, more dependency confusion.
- pykern is  an application support framework. If we have a dep on scipy (and then all the transitive deps that come with that) then we are depending on things that applications we support may depend on. Applications we are going to support will probably depend on scipy/numpy. Who knows what version. Seems like we are opening ourselves up to version conflicts.
- I think developing something where scientist can put pykern like things but more relevant for their domain makes more sense. Maybe a new package (ex rssci)? They could follow different coding standards, add whatever deps are relevant, etc.

e-carlin: Another point, maybe they want to distribute on pypi which we don't currently do for pykern. If we were in different packages then that would be no problem.

bruhwiler:
I agree that a dependency of pykern on scipy is not a good idea.

robnagler:
Would it make sense to create rsmath.const? I would think there would be other shareable algorithms.

bruhwiler:
A python repo rsmath seems like a reasonable solution.
It would be very sparse at first, but I agree that it would be a good place to begin collecting shareable algorithms.

PR #162, opened 2022-04-21
Fix #161: pksetup assert versions

### LINKED ISSUE #161: assert versions in pksetup
setuptools is often broken so assert the versions we expect in pksetup.py so we can fail in a clean way not like https://github.com/radiasoft/devops/issues/265

gurhar1133:
* _assert_package_versions() ensures setuptools version is correct
* Some imports changed in pkcli/test.py so that unknown options in pykern test invocations throws correct error

### DIFF
--- a/pykern/pksetup.py
+++ b/pykern/pksetup.py
@@ -48,7 +48,6 @@ import locale
 import os
 import os.path
 import pkg_resources
-import packaging.version
 import re
 import setuptools
 import setuptools.command.sdist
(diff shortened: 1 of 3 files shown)


pykern/pksetup.py
```diff
@@ -342,7 +342,18 @@ def main():
...
+        """Raise assertion if another module has installed incompatible version of setuptools
+        """
+        from packaging.version import Version
+
+        assert Version(setuptools.__version__) < Version('57'), \
+            'Some package installed a newer version of setup tools that does not work with pykern.pksetup'
+
     name = kwargs['name']
+    if name != 'pykern':
+        # POSIT: setup.py has 'setuptools<57'
```
robnagler, line 355: This comment should be moved to _assert_package_versions docstring.

robnagler (changes requested):

pykern/pksetup.py
gurhar1133: Just move it into the docstring like this
```python
 def _assert_package_versions():
        """Raise assertion if another module has installed incompatible version of setuptools
            POSIT: setup.py has 'setuptools<57'
        """
```
or replace the docstring? ei.
```python
 def _assert_package_versions():
        """POSIT: setup.py has 'setuptools<57'"""
```

robnagler: Move the string and add some space and formatting:

```
"""Raise assertion if another module has installed incompatible version of setuptools

POSIT: setup.py has 'setuptools<57'
""""
```

gurhar1133: You got it 👍

gurhar1133:
Made the change @robnagler

robnagler (approved):

Issue #167, opened 2022-05-11
Update wiki/TestingScientificCodes


bruhwiler:
According to the wiki, 'pytest' should be replaced by 'pykern pytest'
https://github.com/radiasoft/pykern/wiki/TestingScientificCodes

However, 'pykern pytest' seems to call py.test
which seems to have been deprecated to just invoke pytest,
https://github.com/radiasoft/pykern/blob/af25f23c878953e67fddd99912c138d5ad2591da/pykern/pkcli/pytest.py#L11-L13

robnagler:
The correct command is `pykern test`. The documentation is out of date in a number of ways.

robnagler:
@gurhar1133  not a rush, but this needs quite a bit of cleanup. pkmath is pkexample. Remove pkcli/pytest.py. Update for qualified imports. Etc.

robnagler:
You can also say `pykern test .` or `pykern test some_test.py other_test.py`.

PR #180, opened 2022-05-18
Fix #168: recognize if in test dir

### LINKED ISSUE #168: pykern test should recognize it is in the test directory
See #167

gurhar1133:
Now you can run `pykern test` in `tests/` without getting `error: no tests found`

See #167

### DIFF
--- a/pykern/pkcli/test.py
+++ b/pykern/pkcli/test.py
@@ -100,7 +100,6 @@ def _args(tests):
 
 def _find(paths):
     from pykern import pkio
-
     import re
 
     i = re.compile(r'(?:_work|_data)/')
--- a/pykern/pkcli/test.py
+++ b/pykern/pkcli/test.py
@@ -125,8 +125,4 @@ def _resolve_test_paths(paths, current_dir):
         if p.basename != _SUITE_D:
             p = _SUITE_D
         paths = (p,)
-
-    pkdp('\n\n\n cwd')
-    pkdp('\n\n\n paths: {} \n\n\n', paths)
-
     return paths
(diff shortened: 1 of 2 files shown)


pykern/pkcli/test.py
```diff
@@ -104,7 +104,12 @@ def _find(paths):
     i = re.compile(r'(?:_work|_data)/')
     res = []
     cwd = pkio.py_path()
-    for t in paths or ('tests',):
+    d = paths
```
robnagler, line 107: Change the condition order so that it can accept arguments in the tests directory (there should be a test for this):

```py
if not paths:
     p = pkio.py_path()
     if p.basename != _SUITE_D
         p = _SUITE_D
     paths = (p,)
for t in paths:
```

robnagler (changes requested):

pykern/pkcli/test.py
gurhar1133: Nice. I was wondering about how to best structure those conditionals

gurhar1133:
Made changes and added test @robnagler lmk what you think

pykern/pkcli/test.py
robnagler: This temporary is not necessary. `paths` is a variable taht can be falsey, and if it is, then you are setting it to a default value. The extra variable `d` just adds unnecessary complexity. Similarly for `cwd` which is unnecessary.

robnagler (changes requested):

pykern/pkcli/test.py
gurhar1133:
```python
d = paths
``` 
is outdated. Why is cwd unnecessary?

robnagler: I see now, never mind.

robnagler (approved):
There could be more work on this, but it's not that important.

PR #181, opened 2022-05-20
fix radiasoft/devops#252: automate test suite runs on prs

rorour:


rorour:
@robnagler 

- the `uses: with: ref:` that we tried yesterday fails with “permission denied” to clone the repo.
- What is the best way to do pip uninstall before pip install? the uninstall command expects named requirements
- one test (`pkunit_test.py`) is failing with missing dependency despite the `pip install -e .`
- to cd into the repo, there doesn’t seem to be a variable provided by github with just the repo name. $GITHUB_REPOSITORY is rorour/pykern, and I need to be able to cd pykern

robnagler:
I ended up debugging this quite a bit. /rsentry wasn't working, because docker runs at root. The pull checkout is working without that extra arg. I have a working test here: https://github.com/robnagler/ci-play/runs/6540935013?check_suite_focus=true

All the business logic is in the [ci-pull-request](https://github.com/radiasoft/download/blob/d4e3db514093768b5fa700333fbda673cf04626d/installers/ci-pull-request/radiasoft-download.sh) curl installer. The [ci-pull-request.yml](https://github.com/robnagler/ci-play/blob/8765e40d1de8fb258580a01151a5733df8848445/.github/workflows/ci-pull-request.yml) is minimal.

This could be generalized. I also backed off the renaming of environment variables. This is only necessary if we are using them in code.

robnagler (approved):
I think we are ready to go.

PR #196, opened 2022-06-07
Fix #191: black fmt module

### LINKED ISSUE #191: pykern.pkcli.fmt module
Add a pykern.pkcli.fmt with functions:

- edit- runs the formatter (black) on a directory or list of files and edits the files in place
- diff - runs edit and then runs git diff on the files. If there are no differences, returns true (exit(0)).

gurhar1133:
using `pksubprocess.check_call_with_signals()` to invoke black. 
Tests for formatting a single file or a dir as well as diff


gurhar1133:
@robnagler not sure I understand the check failure here. Does it mean that black needs to be included in requirements for pykern somewhere?

robnagler:
Yes, and that means our checks are working. :)

black is a Python module.

gurhar1133:
@robnagler fixed by adding black to setup.py

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,21 @@
...
+
+:copyright: Copyright (c) 2013-2018 Bivio Software, Inc.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkdebug import pkdp
+import pykern.pksubprocess
+
+
+def edit(path):
+    cmd = ['black', f'{path}']
```
robnagler, line 12: temporary variable unnecessary.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,52 @@
...
+    actual_path = pkunit.work_dir().join('file1.py')
+    fmt.edit(actual_path)
+    pkunit.file_eq(
+        expect_path=pkunit.data_dir().join('file1_expect.py'),
+        actual_path=actual_path
+    )
+    pkunit.data_dir().join('fmt_dir').copy(pkunit.work_dir().join('fmt_dir'))
+    actual_path = pkunit.work_dir().join('fmt_dir')
+    fmt.edit(actual_path)
+    for f in pkio.walk_tree(actual_path):
```
robnagler, line 26: use case_dirs :) See for example xlsx_test.py

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,52 @@
...
+        pkunit.file_eq(
+            expect_path=pkunit.data_dir().join(f'fmt_dir_expect/{f.basename}'),
+            actual_path=f
+        )
+
```
robnagler, line 31: too many blank lines. What does black do with this?

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+u"""test fmt
+
+:copyright: Copyright (c) 2019 Bivio Software, Inc.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
```
robnagler, line 7: Not used.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+u"""test fmt
```
robnagler, line 2: black will remove the "u"

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,21 @@
...
+u"""run black formatter on file/dir
+
+:copyright: Copyright (c) 2013-2018 Bivio Software, Inc.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkdebug import pkdp
+import pykern.pksubprocess
+
+
+def edit(path):
```
robnagler, line 11: docstring, please. I don't know if I like edit. I think run is maybe better.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+u"""run black formatter on file/dir
+
+:copyright: Copyright (c) 2013-2018 Bivio Software, Inc.  All Rights Reserved.
```
robnagler, line 4: Definitely copied the copyright from the wrong place. :) If this was copied from a RadiaSoft repo, then fix the original to be an RS copyright with the correct dates.

In emacs, we say C-c C-t fill a template so we don't have to copy and paste from other code.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,21 @@
...
+import pykern.pksubprocess
+
+
+def edit(path):
+    cmd = ['black', f'{path}']
+    pykern.pksubprocess.check_call_with_signals(cmd)
+
+
+def diff(path, path_expect=None):
+    cmd = ['git', 'diff', '--no-index', f'{path}']
```
robnagler, line 17: This should be below the call to edit. Keep definition and use together. However, I think you could actually use `black --diff --no-color`, or even `black --check` , which would be sufficient. Maybe this should just be called fmt.diff and fmt.check to parallel the names black uses.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,21 @@
...
+from pykern.pkdebug import pkdp
+import pykern.pksubprocess
+
+
+def edit(path):
+    cmd = ['black', f'{path}']
+    pykern.pksubprocess.check_call_with_signals(cmd)
+
+
+def diff(path, path_expect=None):
```
robnagler, line 16: Not sure what path_expect means. I think expect means what we expect it to be, that is the reference value. I don't see why there would need to be a reference value passed to this function.

tests/pkcli/fmt_data/file1_expect.py
```diff
@@ -0,0 +1,21 @@
+def bad_foo3(
```
robnagler, line 1: Tests serve as examples and documentation. In this case, it would be nice to see what the formatter will do with different cases instead of the same one.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+:copyright: Copyright (c) 2019 Bivio Software, Inc.  All Rights Reserved.
```
robnagler, line 4: Incorrect copyright.

robnagler (changes requested):

tests/pkcli/fmt_test.py
gurhar1133: Not sure what you mean here?

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_diff2():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    pkunit.data_dir().join('file1.py').copy(pkunit.empty_work_dir().join('file1.py'))
+    actual_path = pkunit.work_dir().join('file1.py')
+    with pkunit.pkexcept(RuntimeError):
```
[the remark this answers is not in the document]
gurhar1133: Should we create a separate issue for addressing intended deviance in case_dirs?

pykern/pkcli/fmt.py
gurhar1133: How about `_diff()` is called if `--check` detects diff and the public method is `check_diff()` ?

tests/pkcli/fmt_test.py
robnagler: There was an extra blank line (3 instead of 2). I was curious if you ran black on this file, would it eliminate the extra line?

tests/pkcli/fmt_test.py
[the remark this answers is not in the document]
robnagler: Yes, good idea. This PR will depend on it.

pykern/pkcli/fmt.py
robnagler: The purpose of check is to be used only in automation. If there's a problem, the solution is to just run fmt, not to see the differences. I realized that knowing the differences wasn't interesting, since formatter will fix everything up. The bug is that the developer committed without running the formatter first.

tests/pkcli/fmt_test.py
gurhar1133: Oh. Let me check

pykern/pkcli/fmt.py
gurhar1133: Ok, so separate `check()` and `diff()` functions?

tests/pkcli/fmt_data/file1_expect.py
gurhar1133: @robnagler made different cases lmk if more things you would like to see black format

pykern/pkcli/fmt.py
robnagler: Yes.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+        'black',
+        '--diff',
+        '--no-color',
+        f'{path}'
+    ])
+
+
+
+def check(path):
+    '''Returns True if there would be diff else return False'''
```
robnagler, line 32: Args:

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+
+:copyright: Copyright (c) 2015 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkdebug import pkdp
+import pykern.pksubprocess
+import pykern.pkunit
+
+def run(path):
+    '''
```
robnagler, line 12: Black prefers double quotes. RS prefers text on one line (black doesn't care). Note the syntax of Args:

```py
"""Run black formatter on `path`

Args:
    path (object): string or py.path to file or directory
```

I used (object) here, because we don't care as long as it can be stringified.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+        Run black formatter on path
+
+        Args:
+            path - string path to file or dir
+    '''
+    pykern.pksubprocess.check_call_with_signals(['black', f'{path}'])
```
robnagler, line 18: Black is unnecessarily noisy and emojiful which doesn't necessarily render in all environments (log files) so I think I would prefer `--quiet`. If people start complaining about that, we can add a configuration parameter.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+def diff(path):
+    pykern.pksubprocess.check_call_with_signals([
+        'black',
+        '--diff',
+        '--no-color',
+        f'{path}'
+    ])
+
+
+
```
robnagler, line 30: Black prefers 2 lines so run the black formatter on fmt.py before committing, because there are three blank lines.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+    '''
+        Run black formatter on path
+
+        Args:
+            path - string path to file or dir
+    '''
+    pykern.pksubprocess.check_call_with_signals(['black', f'{path}'])
+
+
+def diff(path):
```
robnagler, line 21: docstring.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+    pykern.pksubprocess.check_call_with_signals([
+        'black',
+        '--diff',
```
robnagler, line 24: Add `--quiet` which does the right thing (shows the diff, but not the emojis or statements like "unchanged")

diff should return false or true if there are differences. This  can be accomplished by `black --check --diff`.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+        '--no-color',
+        f'{path}'
+    ])
+
+
+
+def check(path):
+    '''Returns True if there would be diff else return False'''
+    try:
+        pykern.pksubprocess.check_call_with_signals([
```
robnagler, line 34: rule of three says to factor out `pykern.pksubprocess.check_call_with_signals(["black"` as it is repeated.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,41 @@
...
+    for d in pkunit.case_dirs():
+        fmt.run(d)
+
+
+def test_diff():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    fmt.diff(pkunit.data_dir().join('file1.py'))
```
robnagler, line 33: I think the output of `diff` should be checked. You can capture output in pytest. I think there are other examples of this. This is where you can use file_eq.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_check():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    pkunit.pkok(fmt.check(pkunit.data_dir().join('file1.py')), 'expect check to return True')
```
robnagler, line 41: `pkeq` is better and eliminates need for a comment string.

pykern/pkcli/fmt.py
```diff
@@ -0,0 +1,42 @@
...
+            'black',
+            '--check',
+            f'{path}'
+        ])
+    except RuntimeError as e:
+        if str(e) == 'error exit(1)':
```
robnagler, line 40: I don't like the str, but it's ok given the interface. Returning True is not how you return an error (check fails). You have to call command_error. Also, call `raise` when you don't get what you expect (another exception):

```py
except RuntimeError as e:
        if  str(e) == "error exit(1)":
            pykern.pkcli.command_error('path={} needs to be formatted', path)
        raise
```

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,41 @@
...
+from pykern.pkdebug import pkdp
+import pytest
+
+
+def test_run():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    pkunit.data_dir().join("file1.py").copy(pkunit.empty_work_dir().join("file1.py"))
```
robnagler, line 17: This code should not be necessary if case_dirs is used.

robnagler (changes requested):

pykern/pkcli/fmt.py
gurhar1133: I think you just coined a new term in "emojiful"

pykern/pkcli/fmt.py
gurhar1133: This function just checks if their is diff. So do we really want to raise an error every time there is diff ('error exit(1)' being the one `black check ...` raises) or do we just want to use check to return True or False that their will be diff?

tests/pkcli/fmt_data/file1_diff_expect.txt
```diff
@@ -0,0 +1,26 @@
...
++    ever,
++    use,
++    it,
++    has,
++    way,
++    too,
++    many,
++    arguments,
++):
++    print("force_arg_per_line")
```
robnagler, line 26: why the print?

tests/pkcli/fmt_data/fmt_dir.in/y.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+
+
+    y= 100
+    z= 200
+    p, q = 1, 2
+
+
+    print(y+z+p+q)
```
robnagler, line 20: why the print?

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,49 @@
...
+import pytest
+
+
+def test_run():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    for d in pkunit.case_dirs():
+        if d == '/home/vagrant/src/radiasoft/pykern/tests/pkcli/fmt_work/single_file':
```
robnagler, line 19: No hardwired paths and I don't think this is necessary. Just fmt.run on the directory.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,49 @@
...
+        fmt.run(d)
+
+
+def test_diff():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pksubprocess import check_call_with_signals
+
+    actual_path = pkunit.work_dir().join('file1_diff_expect.py')
+    check_call_with_signals([
```
robnagler, line 30: You should be able to call directly.

robnagler (changes requested):

tests/pkcli/fmt_data/fmt_dir.in/y.py
gurhar1133: @robnagler The print is random, for the code to document the test case should I call `foo(y+z+p+q)` instead to show that it is just placeholder?

robnagler: Yes, prints are only useful if you want to test the output. Rather just do something that doesn't cause output.

pykern/pkunit.py
```diff
@@ -71,7 +71,7 @@ def assert_object_with_json(basename, actual, ):
     pkeq(expect, actual, 'diff {} {}', e, a)
 
 
-def case_dirs():
+def case_dirs(group_prefix=''):
```
robnagler, line 74: Add to docstring

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,47 @@
...
+def _cases_with_prefix_except(prefix, fn):
+    from pykern import pkunit
+    from pykern import pkio
+
+    for d in pkunit.case_dirs(prefix):
+        try:
+            fn(d)
+            pkio.write_text('res', 'check ok\n')
+        except Exception as e:
+            pkio.write_text('res', f'check exception=failed\n')
```
robnagler, line 47: `pkio.write_text('res', f'check exception={e}\n')` so we see the error and know what it should/shouldn't be.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,47 @@
...
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    _cases_with_prefix_except('diff', fmt.diff)
+
+
+def test_run():
+    from pykern import pkunit
+    from pykern import pkio
```
robnagler, line 31: unnecessary import

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,47 @@
...
+def test_check():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    _cases_with_prefix_except('check', fmt.check)
+
+
+def test_diff():
+    from pykern import pkunit
```
robnagler, line 22: no imports necessary

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,47 @@
...
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from multiprocessing.dummy import active_children
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdp
+import pytest
+
+
+def test_check():
+    from pykern import pkunit
```
robnagler, line 14: no imports necessary

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,47 @@
...
+def test_run():
+    from pykern import pkunit
+    from pykern import pkio
+    from pykern.pkcli import fmt
+
+    for d in pkunit.case_dirs('fmt_dir'):
+        fmt.run(d)
+
+
+def _cases_with_prefix_except(prefix, fn):
```
robnagler, line 38: Nice refactoring.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,47 @@
...
+    from pykern import pkunit
+    from pykern import pkio
+
+    for d in pkunit.case_dirs(prefix):
+        try:
+            fn(d)
+            pkio.write_text('res', 'check ok\n')
```
robnagler, line 45: Refactor as follows:

```py
res = 'check ok'
```

tests/pkcli/fmt_test.py
robnagler, line 47: Refactor as follows:

```py
    f'check exception={e}'
```

Note the `{e}` so we see the error and know what it should/shouldn't be.

Then add a line at the end (unindented):

```py
pkio.write_text('res', res + "\n")
```

robnagler (changes requested):

gurhar1133:
@robnagler 
This:
```python
except Exception as e:
    res = f'check exception={e}\n'
```
gives test diff for `test_check()`:
```
< check exception=failed
---
> check exception=path=/home/vagrant/src/radiasoft/pykern/tests/pkcli/fmt_work/check1 needs to be formatted
```

The path will be different in github auto testing and will therefore fail. Should I do something like:
```python
except Exception as e:
    m = str(e).replace(str(d), '')
    res = f'check exception={m}\n'
```
?

robnagler:
I didn't know what it output. Yes, a string replace would be good. You don't need a temporary variable or you could put it in a function if you want to give it a long name.

pykern/pkunit.py
```diff
@@ -83,6 +83,11 @@ def case_dirs(group_prefix=''):
     of these expect files is copmared to the corresponding `work_dir`
     actual file using `file_eq`.
 
+    If you want to only use cases from some specific `/<case-name>.in`
```
robnagler, line 86: remove "/" from paths.

pykern/pkunit.py
```diff
@@ -83,6 +83,11 @@ def case_dirs(group_prefix=''):
...
+    subdir, and not all `/*.in` subdirs, you can pass a `group_prefix`
+    default parameter value ('' by default) to `case_dirs()`. This will
+    perform the regular operations but only on `/<case-name>.in`.
+
```
robnagler, line 90: Add an `Args:` section. See example in [pkjson](https://github.com/radiasoft/pykern/blob/3c0eda21925a2df323299b6658d5ac5c216335a0/pykern/pkjson.py#L29) for how to handle optional args.

robnagler (changes requested):
Almost there!

gurhar1133:
@robnagler made changes, LMK your thoughts :)

pykern/pkunit.py
robnagler: Still don't see this.

tests/pkcli/fmt_test.py
```diff
@@ -0,0 +1,43 @@
...
+    from pykern import pkunit
+    from pykern import pkio
+
+    for d in pkunit.case_dirs(prefix):
+        try:
+            fn(d)
+            res = 'check ok\n'
+        except Exception as e:
+            res = f'check exception={e}\n'
+        pkio.write_text('res', res)
```
robnagler, line 43: Need to parse out `d` from the exception, e.g. `res.replace(str(d), '')`

robnagler (changes requested):

pykern/pkunit.py
gurhar1133: It's in most recent commit

gurhar1133:
@robnagler i think your comments arent on the most recent couple of commits.

robnagler (approved):

PR #206, opened 2022-06-20
202 ci check prints

gurhar1133:
Added ci.check_prints
Test for detect a file and not detect a file
Check prints is called in ci.run


pykern/pkcli/ci.py
```diff
@@ -5,17 +5,47 @@
...
 """
 from pykern.pkdebug import pkdp, pkdlog
+from pykern import pkunit
+from pykern.pkcli import test
+from pykern import pksetup
+import re
+
+
+_FILE_TYPE = re.compile(r'.py$')
+_EXCLUDE_FILES = re.compile( f"/{test.SUITE_D}/.*{pkunit.DATA_DIR_SUFFIX}/|/{test.SUITE_D}/.*_work/|/{pksetup.PACKAGE_DATA}/|pkdebug")
```
rorour, line 15: There's an extra space before the f, also I see double quotes used in a few places in this pr.

tests/pkcli/ci_test.py
```diff
@@ -5,17 +5,31 @@
...
         pkio.write_text('res', res)
+
+
+def test_check_prints():
+    from pykern.pkcli import ci, test
+    from pykern import pkio
+    from pykern import pkunit
+    from pykern import pksetup
+
+    r = re.compile( f"/{test.SUITE_D}/.*{pkunit.DATA_DIR_SUFFIX}/|/{pksetup.PACKAGE_DATA}/|pkdebug")
```
rorour, line 32: This regex is similar to the one used on l15 of ci.py. It could be a good idea to link them (possibly by using `ci._EXCLUDE_FILES` here) in case we ever need to change both of them.

rorour (commented):
@robnagler What is the benefit of using an alias instead of calling print() directly? Just another layer to ensure we don't leave any prints that we don't need?

tests/pkcli/ci_data/check_prints1.in/x.py
```diff
@@ -0,0 +1,4 @@
+def check_prints():   
+    from pykern.pkdebug import pkdp
+    pkdp("this pkdp should be found")
```
robnagler, line 3: blank line above.

tests/pkcli/ci_data/check_prints2.in/x.py
```diff
@@ -0,0 +1,3 @@
+def foo():
+    from pykern.pkdebug import pkdp
+    mention = (pkdp, print)
```
robnagler, line 3: blank line, and test mentioning in a comment, e.g. 
#pkdp("not found")

pykern/pkcli/ci.py
```diff
@@ -5,17 +5,47 @@
...
+    for f in pkio.walk_tree(pkio.py_path(), _FILE_TYPE):
+        if re.search(exclude, str(f)):
+            continue
+        s = pkio.read_text(str(f))
+        for i, l in enumerate(s.split('\n')):
+            if re.search(_PRINT, l):
+                res += f'{f.basename} pkdp/print on line: {i + 1} :-> {l}\n'
+    if res:
+        pkconst.builtin_print(res)
+        raise AssertionError(f'\n{res}\nChecks fail due to pkdp/print present')
```
robnagler, line 38: Too many blank lines in the output. It's unnecessary to have a leading blank line. You don't know what comes before.  The "Checks fail due to ..." is implicit, so just output `res`. This command will be used in a context that will cause this command to be the last output. Thing "bigger" when designing error messages: where will it be seen and how will it be used? For a programmer running it, they'll see a failed output in a compile/test window. For CI, the test will cause the CI to stop. This is why "emojiful" output is useless in most cases. It looks pretty in one offs, but in the larger context it just creates noise for the developer to wade through. A good IDE will let you know a test/command fails.

tests/pkcli/ci_data/check_prints1.in/x.py
robnagler: Add the case of a pkdp and the module level also test for print. The function name "check_prints" doesn't really name what's going on. What would be a better name? The import statement can be at the global level, because this is a test script.

tests/pkcli/ci_test.py
robnagler, line 32: This bypasses a big part of the code. How would you test exclude but still allow this test to function? (Hint: pkunit.module_under_test, which maybe could be better encapsulated.)

robnagler (approved):

pykern/pkcli/ci.py
```diff
@@ -5,17 +5,47 @@
...
+from pykern import pksetup
+import re
+
+
+_FILE_TYPE = re.compile(r'.py$')
+_EXCLUDE_FILES = re.compile( f"/{test.SUITE_D}/.*{pkunit.DATA_DIR_SUFFIX}/|/{test.SUITE_D}/.*_work/|/{pksetup.PACKAGE_DATA}/|pkdebug")
+_PRINT = re.compile('\s(pkdp)\(|\s(print)\(')
+
+
+def check_prints(exclude=_EXCLUDE_FILES):
```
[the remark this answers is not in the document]
gurhar1133: The idea with exclude was that in testing we actually dont want to ignore _work. Otherwise we cant test the function.

tests/pkcli/ci_test.py
gurhar1133: I'm not sure I follow you. I looked at the module_under_test in pykern code and it was not clear to me what that is doing.

pykern/pkcli/ci.py
```diff
@@ -5,17 +5,52 @@
...
+    res = []
+    for f in pkio.walk_tree(pkio.py_path(), _FILE_TYPE):
+        if re.search(_EXCLUDE_FILES, str(f)):
+            if not (pkunit.is_test_run() and '/ci_work/' in str(f)):
+                continue
+            pkdlog('special case /ci_work/ path={}', f)
+        for i, l in enumerate(pkio.read_text(f).split('\n'), start=1):
+            if re.search(_PRINT, l):
+                res.append(f'{f.basename}:{i} {l}')
+    res = '\n'.join(res)
```
robnagler, line 41: Subtle, but better:

```py
pkcli.command_error('\n'.join(res))
```

pykern/pksetup.py
```diff
@@ -334,7 +334,7 @@ def setup(**kwargs):
         The file ``pykern_console.py`` might contain::
 
             def main():
-                print('hello world')
+                builtin_print('hello world')
```
robnagler, line 337: This is an odd one. Maybe better to change it to:

```py
def main():
    return 2 + 2
```

tests/pkcli/ci_test.py
```diff
@@ -5,17 +5,34 @@
...
 import pytest
 
+
 def test_run():
     from pykern.pkcli import ci
     from pykern import pkio
     from pykern import pkunit
 
-    for d in pkunit.case_dirs():
+    for d in pkunit.case_dirs("case"):
```
robnagler, line 17: change "case" to "run".

tests/pkcli/ci_test.py
```diff
@@ -5,17 +5,34 @@
...
         try:
```
robnagler, line 18: can you use pkexcept_to_file()

robnagler (changes requested):
Minor nits.

gurhar1133:
@robnagler made the changes 👍

robnagler (approved):

PR #210, opened 2022-06-22
fix #203 remove script dir from sys.path

### LINKED ISSUE #203: Remove script dir from sys.path?
Try this from any directory (not in sys.path):

```
echo 'import csv;csv.reader' > csv.py
python csv.py
```

It will give this error: `AttributeError: module 'csv' has no attribute 'reader'`

@garsuga ran into this error, and I was confused by it. I now see that sys.path gets PWD inserted. This broke @garsuga's code that we had to stop using the program, because sirepo.numpy exists. (The program cd'd to the module being debugged, which is a separate issue.)

I think we should consider dropping the first element in sys.path in pykern.pkcli. (We should also add a fixup to projex that imports everything unqualified to reduce [but not eliminate] collisions.)

This has been [known in Unix since 1984](https://people.engr.ncsu.edu/gjin2/Classes/246/Spring2019/Security.pdf). 

Go recently [added something to prevent this](https://github.com/golang/go/issues/43724). I think this is a breaking change.

Since all processes are started with pykern.pkcli.main (except sirepo_jupyterhub which could be changed), it could strip '.' from the first element of sys.path. I think this would avoid problems and be more secure.

robnagler:
- Require test?
   - yes
- Security implications?
    - Better
- UI changes reviewed?
    - None
- Backwards compatible?
    - No, but should be ok
- Require doc changes?
    - No
-  Add [DesignHints](https://github.com/radiasoft/pykern/wiki/DesignHints)?
    - No


pykern/pkcli/__init__.py
```diff
@@ -333,3 +338,22 @@ def _module_name(path_list):
...
+    """
+    global _fix_sys_path_done
+
+    if _fix_sys_path_done:
+        return
+    _fix_sys_path_done = True
+    if not (sys.argv and sys.path):
+        # Not enough information
+        return
+    d = os.path.dirname(os.path.realpath(sys.argv[0]))
```
e-carlin, line 357: Any reason for this check instead of `if sys.path[0] != ''; sys.path.pop(0)`

https://docs.python.org/3.7/library/sys.html#sys.path

e-carlin (approved):

pykern/pkcli/__init__.py
robnagler: That's a different case which does not exist if pkcli.main is used. We always have a script. This case is really not an issue:

```sh
$ python
> import pykern.pkcli
> pykern.pkcli.main('pykern')
```

We also don't have the '-m' case, which is when you pass in a module.

PR #218, opened 2022-06-24
Fix #215: ingore dot prefix

### LINKED ISSUE #215: pkcli.ci ignore all dot-prefix dirs
If there's a `.tox` directory, check_prints will fail. I don't think check_prints or run should look in dot prefixed files at all. There's a lot of them, and they don't contain anything we want to check regularly.

gurhar1133:
Dot prefixes are now ignored
ci1_test.py tests dot prefixes, _data and package_data


tests/pkcli/ci1_test.py
```diff
@@ -0,0 +1,31 @@
...
+import re
+import pytest
+
+
+def test_check_prints_dot_prefix():
+    from pykern.pkcli import ci, test
+    from pykern import pkunit
+    from pykern import pksetup
+
+    ci._EXCLUDE_FILES = re.compile(
```
robnagler, line 18: This should be an edited value, that is, start with the existing value and edit out the _WORK_DIR_SUFFIX VALUE. Useful fact: re.compile('my regex').pattern == 'my regex'.

tests/pkcli/ci1_test.py
```diff
@@ -0,0 +1,31 @@
...
+        f"/{test.SUITE_D}/.*{pkunit.DATA_DIR_SUFFIX}/"
+        + f"|/{pksetup.PACKAGE_DATA}/"
+        + r"|pkdebug[^/]*\.py$"
+        + r"|\.*/"
+    )
+
+    for d in pkunit.case_dirs("ignore_dot_prefix"):
```
robnagler, line 25: I think having separate case_dirs causes confusion. Just one directory with all the different cases with mistakes in them that all should be ignored.

pykern/pkcli/ci.py
```diff
@@ -16,6 +16,7 @@
     f"/{test.SUITE_D}/.*(?:{pkunit.DATA_DIR_SUFFIX}|{pkunit.WORK_DIR_SUFFIX})/"
     + f"|/{pksetup.PACKAGE_DATA}/"
     + r"|pkdebug[^/]*\.py$"
+    + r"|\.*/"
```
robnagler, line 19: This is incorrect. Must start with a dot, not end with a dot.

robnagler (changes requested):

tests/pkcli/ci1_test.py
gurhar1133: Wont having a separate test.py necessitate a separate _data dir?

robnagler: Have one test (ci1_test) with one case dir (ci1_data/1.in).

gurhar1133:
@robnagler checks pass now, lmk if you have any more change requests :)

tests/pkcli/ci1_test.py
```diff
@@ -0,0 +1,27 @@
...
+import pytest
+
+
+def test_check_prints_dot_prefix():
+    from pykern.pkcli import ci, test
+    from pykern import pkunit
+
+    ci._EXCLUDE_FILES = re.compile(
+        re.sub(
+            f"/{test.SUITE_D}/.*(?:{pkunit.DATA_DIR_SUFFIX}|{pkunit.WORK_DIR_SUFFIX})/",
```
robnagler, line 19: Don't use re.sub when there are specials without escaping the pattern, use string.replace. Also the search pattern can be greatly simplified and probably should be asserted, so:

```py
def _unexclude_work_dir():
    p = ci._EXCLUDE_FILES.pattern.replace(f"|{pkunit.WORK_DIR_SUFFIX}", "")
    assert p != ci._EXCLUDE_FILES.pattern
    ci._EXCLUDE_FILES = re.compile(p, flags=ci._EXCLUDE_FILES.flags)
```

pykern/pkcli/ci.py
```diff
@@ -16,6 +16,7 @@
     f"/{test.SUITE_D}/.*(?:{pkunit.DATA_DIR_SUFFIX}|{pkunit.WORK_DIR_SUFFIX})/"
     + f"|/{pksetup.PACKAGE_DATA}/"
     + r"|pkdebug[^/]*\.py$"
+    + r"|.*/\.[^/]*"
```
robnagler, line 19: You only need `r"|/\."`.

tests/pkcli/ci1_data/1.in/package_data/x.py
```diff
@@ -0,0 +1 @@
+print("this should be ingore since package_data")
```
robnagler, line 1: typo

robnagler (changes requested):

tests/pkcli/ci1_test.py
```diff
@@ -14,14 +14,16 @@ def test_check_prints_dot_prefix():
...
 
-    ci._EXCLUDE_FILES = re.compile(
-        re.sub(
-            f"/{test.SUITE_D}/.*(?:{pkunit.DATA_DIR_SUFFIX}|{pkunit.WORK_DIR_SUFFIX})/",
-            f"/{test.SUITE_D}/.*{pkunit.DATA_DIR_SUFFIX}/",
-            ci._EXCLUDE_FILES.pattern
-        )
-    )
-
+    _unexclude_work_dir()
```
robnagler, line 17: I don't think this should be necessary with relative paths, because pwd will be inside _work.

pykern/pkcli/ci.py
```diff
@@ -31,14 +31,18 @@ def check_prints():
     from pykern import pkcli
 
     res = []
-    for f in pkio.walk_tree(pkio.py_path(), _FILE_TYPE):
+    p = pkio.py_path()
+    for f in pkio.walk_tree(p, _FILE_TYPE):
+        f = "./" + p.bestrelpath(f)
         if re.search(_EXCLUDE_FILES, str(f)):
```
robnagler, line 37: There are two primary cases:

1. We are running from top of repo. This is what "ci" is supposed to do.
2. Ad hoc testing, iwc arguments are passed (currently not supported)

I'm focusing on case (1). We can assert we are running from the top of the repo (there should be a setup.py in the pwd, for example). Given that, the regex can make more assumptions:

```py
    f"^{test.SUITE_D}/.*(?:{pkunit.DATA_DIR_SUFFIX}|{pkunit.WORK_DIR_SUFFIX})/"
    + f"|^\\w+/{pksetup.PACKAGE_DATA}/"
    + r"|pkdebug[^/]*\.py$"
    + r"|(?:^|/)\."
```

In the ad hoc case (2) we can just accept file(s) as arguments and check prints on those. I don't think we need to support that case right now, though.

Also "f" is  a string already so "str()" is unnecessary.

pykern/pkcli/ci.py
```diff
@@ -31,14 +31,18 @@ def check_prints():
     from pykern import pkcli
 
     res = []
-    for f in pkio.walk_tree(pkio.py_path(), _FILE_TYPE):
+    p = pkio.py_path()
+    for f in pkio.walk_tree(p, _FILE_TYPE):
+        f = "./" + p.bestrelpath(f)
```
robnagler, line 36: Don't need "./"

pykern/pkcli/ci.py
```diff
@@ -31,14 +31,18 @@ def check_prints():
...
         if re.search(_EXCLUDE_FILES, str(f)):
-            if not (pkunit.is_test_run() and "/ci_work/" in str(f)):
+            if not (
```
robnagler, line 38: No longer necessary.

tests/pkcli/ci1_data/1.in/package_data/x.py
```diff
@@ -1 +1 @@
-print("this should be ingore since package_data")
+print("this should be ignore since package_data")
```
robnagler, line 1: So now the test directory needs to be structured like a real repo with a setup.py (can be blank) and package_data under some package name. The same is true of the rest of the files.

tests/pkcli/ci1_test.py
```diff
@@ -14,14 +14,16 @@ def test_check_prints_dot_prefix():
...
     for d in pkunit.case_dirs():
         with pkunit.pkexcept_to_file():
             ci.check_prints()
+
+
+def _unexclude_work_dir():
+    from pykern.pkcli import ci
+    from pykern import pkunit
+
+    p = ci._EXCLUDE_FILES.pattern.replace(f"|{pkunit.WORK_DIR_SUFFIX}", "")
```
robnagler, line 27: No longer necessary.

robnagler (changes requested):

tests/pkcli/ci1_data/1.in/package_data/x.py
gurhar1133: @robnagler made changes with the above dir structure. LMK what you think

robnagler: This is not a repo structure. Look at a repo and see how the above differs from what is there.

gurhar1133: Ok so what about this?
```
 1.in/
    setup.py
    in/
        has_prints.py
        package_data/
            x.py
    tests/
        test_data/
           x.py
   
```

robnagler: x.py could be renamed "not_checked.py" or something so the file names reflect what the test is supposed to be doing.

gurhar1133:
Restructured the repo @robnagler LMK your thoughts

robnagler (approved):

PR #224, opened 2022-06-29
Fix #137 added fconf and PKDict.pkmerge

### LINKED ISSUE #137: Create pykern.fconf
[rsconf parses yaml files](https://github.com/radiasoft/rsconf/blob/df7a4a6fc52f105c8c2b3fc1f44139dbf6cfee11/rsconf/db.py#L87) to produce a single PKDict allowing modular config files, e.g. keeping secrets out of the main config file. It uses jinja2 for named abstractions, loops, vars, etc. This has started to get messy, quite similar to the [Sirepo schemas](https://github.com/radiasoft/sirepo/blob/cf2b5c67cfcb28a92a66ffcdb641ae41b11e0644/sirepo/simulation_db.py#L883) which do not have abstractions, and rely on specialized code to provide aggregation and merging. 

I converted most of our system rsconf yaml files to a new type of syntax that relies on YAML and Python in separate files (thanks @e-carlin for the suggestion). The new yaml files look like:

```yaml
---
self:
  str1: value1
  dict1:
    k1: v1
    k2: [1, 2, 3]
  template1:
    something: 
       - ${arg1}
       - an embedded ${arg2} here

app_stuff:
  goes: here
  simple: str1()
  more:
    dict1(): n/a # this can be anything
    a: b
    c: template1('val1', 'val2')
```

The `self` section is an instance of a PKDict subclass. The `app_stuff` is the "result" used by the application. In the above example, it would produce:

```py
{
    "app_stuff": {
	"goes": "here",
	"more": {
            "a": "b",
            "c": [
                "val1",
		"embeddedval2 here"
            ],
            "k1": "v1",
            "k2": [1, 2, 3]
	},
        "simple": "value1"
    },
}
```

The simple macro language is compatible with YAML and maps well to our system rsconf. There are many simple values and templates that can be shared easily this way.

There is however a need for algorithmic sharing, and that's where the Python comes in:

```py
def public_ip(self, num):
    return f'1.2.3.{num}'

def public_cidr(self):
    return f'{self.public_ip(128)}/25'

def device(self, res, name, ip):
    res[name] = PKDict(ip=ip)
    return res[name]

def devices(self, num, public_num):
    res = PKDict()
    self.device('en0', f'192.186.1.{num}')
    if public_num:
	res.device(
            'en1',
            self.public_ip(public_num),
	).defroute = True
    return PKDict(devices=res)

def host_cores(self, i):
    return self.host_cores_map.get(i, self.host_cores_default())

def blades(self):
    return [self.blade(i) for i in range(1, 16)]

def init_prod_mpi(self, content):
    for k, v in self.host_mpi_user.items():
        r[k] = [self.blade(i) for i in v]
        self._host_mpi.append(r[k])
    self._mpi_worker_prod = PKDict(mpi_worker=PKDict(clusters=r))
    return None

def beta_workers(self):
    res = PKDict()
    for i in self.beta_worker_nums():
        res.update(self.beta_worker(i))
    return res
```

These functions act just like templates, but they can modify their state algorithmically. They can access their current position in the tree, and find their parent(s), e.g. to modify state "around" a use. These "macros" are more powerful than simple textual replacement macros as a result. This can, for example, be used to implement a control structure, similar to Lisp macros.

robnagler:
- [x] Require test?
   - fconf_test.py
- [x] Security implications?
    - no
- [x] UI changes reviewed?
    - design is new and needs discussion
- [x] Backwards compatible?
    - no - removed some py2 support in pkcompat
- [x] Require doc changes?
    - yes
    - no

@mkeilman  I thought this would be interesting way to describe the future schema.
@e-carlin this will be integrated in rsconf
@garsuga this is an example of a DSL that uses existing parsers


pykern/pkcollections.py
```diff
@@ -99,9 +101,77 @@ def pkdel(self, name, default=None):
...
+                f"key={key} type mismatch between (self) base={base} and to_merge={merge}"
+            )
+
+        for k in list(to_merge.keys()):
+            t = to_merge[k]
+            s = self.get(k)
+            if isinstance(s, dict) or isinstance(t, dict):
+                if s is None:
+                    pass
+                elif t is None:
```
e-carlin, line 141: I think this can be collapsed into the first `if` like you do for the list check below

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+loops or other conditions. These are reserved for Python files.
+
+A macro use must be a full text element. It cannot be embedded in
+text, e.g.  ``curl make_uri('sirepo.com')``, that's because macros can
+return data structures. The syntax is pure Python. The arguments have
+to quoted.
+
+More complex macros can be programmed in a Python file::
+
+    import sqrt
```
e-carlin, line 30: import math

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+       - item
+    uri: make_uri('sirepo.com')
+
+Macros in the `fconf_macros` are simple replacements.  They cannot have
+loops or other conditions. These are reserved for Python files.
+
+A macro use must be a full text element. It cannot be embedded in
+text, e.g.  ``curl make_uri('sirepo.com')``, that's because macros can
+return data structures. The syntax is pure Python. The arguments have
+to quoted.
```
e-carlin, line 26: This conflicts with the integer statement below which I found confusing. Are yaml macro args the only ones that have to be quoted?

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+
+The same is true of dicts. For example::
+
+    fconf_macros:
+      base(num):
+        v${num}.radia.run:
+          description: VM ${num}
+
+    all_hosts:
+      base(3):
```
e-carlin, line 88: I didn't know this was valid yaml. Funky.

This example was my "aha" moment. This would be such a pain with jinja due to indenting.

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+    {
+        'a': {'b': {'c': 3}},
+        'd': {'b': {'c': 3}},
+        'e': 3,
+    }
+
+This allows for flexible constants. Those global strings can be used
+in YAML Macros.
+
+YAML files are evaluated before they are merged. However, they are all
```
e-carlin, line 125: I think it would be nice to expand on the order in which files are evaluated and why it is the way it is.

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+import pykern.pkrunpy
+import re
+import pykern.pkyaml
+import pykern.pkcollections
+import pykern.pkio
+
+
+#: parse_files macro expansion pattern
+_TEXT_MACRO_NAME = re.compile(
+    r"^([a-z]\w*)\(((?:\s*[a-z]\w*\s*)?(?:,\s*[a-z]\w*\s*)*)\)$", flags=re.IGNORECASE
```
e-carlin, line 147: If you think there is some way in english to communicate what this regex is testing for then I think that would be good to add. I know I can plug this into regex101.com and fumble my way through it but it really slows down my ability to read code when I come across a regex like this. My regex knowledge has 3x at this job but not well enough I can just look at this for a sec, get what it does, and keep on reading.

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+)
+
+_MACRO_CALL = re.compile(r"^([a-z]\w*)\((.*)\)$", flags=re.IGNORECASE + re.DOTALL)
+
+_ARG_SEP = re.compile(r"\s*,\s*|\s+")
+
+_FVAR = re.compile(r"\$\{(\S+)\}")
```
e-carlin, line 154: I think something more descriptive than "F" would be helpful.

pykern/fconf.py
e-carlin: Maybe this would be better left for the parse_all doc string. Doesn't matter to me.

pykern/fconf.py
```diff
@@ -0,0 +1,517 @@
...
+    def __init__(self, files):
+        self.pkupdate(
+            files=PKDict(py=[], yml=[]),
+            macros=PKDict(),
+        )
+        for f in files:
+            try:
+                self._add_file(f)
+            except Exception:
+                pkdlog("error parsing file={}", f)
```
e-carlin, line 184: What about using [`raise from`](https://stackoverflow.com/a/46091127/5518313) ? I could see myself not seeing the pkdlog above the error.

e-carlin (commented):
This is a really exciting addition! Going to clean up a lot of our rsconf and cool to see how something like this is implemented. Thanks for making this change:)

I'm checkpointing my review here but not done.

mkeilman (commented):
Just a few niggles

pykern/pkcollections.py
```diff
@@ -99,9 +101,77 @@ def pkdel(self, name, default=None):
...
+                    # Just replace, because t's type (None) overrides in case of None.
+                    pass
```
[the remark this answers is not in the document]
robnagler: Thanks. I greatly simplified the logic in 2d18016

pykern/pkcollections.py
```diff
@@ -99,9 +101,77 @@ def pkdel(self, name, default=None):
...
+                    # this test ensures we don't have dup entries in lists.
+                    y = [x for x in self[k] if isinstance(x, collections.abc.Hashable)]
+                    assert len(set(y)) == len(
+                        y
+                    ), f"duplicates in key={k} list values={self[k]}"
+                    continue
+                else:
+                    raise _type_err(k, s, t)
+            elif type(s) != type(t) and not (s is None or t is None):
+                raise _type_err(k, s, t)
```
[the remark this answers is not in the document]
robnagler: That was the problem with the logic. 2d18016 should eliminate your concern.

pykern/pkcollections.py
```diff
@@ -99,9 +101,77 @@ def pkdel(self, name, default=None):
...
+                    continue
+                else:
+                    raise _type_err(k, s, t)
+            elif isinstance(s, list) or isinstance(t, list):
+                if s is None or t is None:
+                    # Just replace, because t overrides type in case of None
+                    pass
+                elif isinstance(s, list) and isinstance(t, list):
+                    # prepend the to_merge values; creates a new list
+                    self[k] = t + s
```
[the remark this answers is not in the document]
mkeilman: I see, I was reading the code change and not the full context

pykern/pkcollections.py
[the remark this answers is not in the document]
mkeilman: 👍

pykern/pkcollections.py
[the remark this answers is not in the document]
mkeilman: 👍 looks good

pykern/fconf.py
robnagler: Thanks! I didn't know about raise from. 6a9b1c0

pykern/pkcollections.py
robnagler: Yes, see https://github.com/radiasoft/pykern/pull/224#discussion_r910031818

pykern/fconf.py
robnagler: It's actually a very interesting construct, because:

```yml
x:
```

generates `{"x": None}` which is used to enforce that macro uses as keys that result in dicts have None as values. This will allow in some future version (if necessary):

```yml
x:
   y: 3
```

where the `{"y": 3}` would be available to the macro implementation to embed (YAML macros) or to evaluate (Python macros). The latter case would allow control structures like real macros.

pykern/fconf.py
robnagler: I can't think of anything. It's an internal name (for now). It's used for both argument and "current value in the eval" expansions. It's used in the code. I tried to clear up with a comment 5ee9801

pykern/pkcollections.py
```diff
@@ -99,9 +101,65 @@ def pkdel(self, name, default=None):
             except KeyError:
                 pass
 
+    def pkmerge(self, to_merge):
+        """Add `to_merge` to `self`
+
+        `self` will have references to `to_merge` after this call.
```
robnagler, line 107: @e-carlin @mkeilman What do you think about destructive copies? I just ran into this being a problem when coding rsconf. Perhaps it could be a flag so that fconf can use it for its purposes.

robnagler: I added make_copy as a parameter in 3f60b12. It makes no sense to make it destructive by default.

mkeilman: I think it's a good idea to give the option; I'm not certain what the default behavior should be though.

mkeilman: Too slow :)

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+        if exc.args is None:
+            exc.args = tuple()
+        if isinstance(exc.args, (tuple, list)):
+            if len(exc.args) == 0:
+                exc.args = (reason,)
+            elif isinstance(exc.args[0], str):
+                x = list(exc.args)
+                x[0] += reason
+                exc.args = tuple(x)
+    # Other cases?
```
e-carlin, line 569: Can you add a comment saying this doesn't have a default case

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+                    content=c,
+                    name=n,
+                    # TODO(robnagler): check for dups
+                    # TODO(robnagler): assert varnames don't begin with fconf_
+                    params=tuple((x for x in _ARG_SEP.split(m.group(2)) if x)),
+                    source=source,
+                ),
+            )
+
+    def _ext_builtins(self, path):
```
e-carlin, line 263: path isn't used

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+        for n, o in inspect.getmembers(obj):
+            if inspect.isfunction(o):
+                m[n] = o
+        return m
+
+
+class _Builtins:
+    @staticmethod
+    def fconf_var(namespace, name):
+        return namespace._evaluator.fconf_var(name)
```
e-carlin, line 283: `_evaluator` should be public if accessed like this

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+        super().__init__(*args, **kwargs)
+        # TEST: duplicate params, invalid params, etc.
+        s = set()
+        for p in self.params:
+            if p in s:
+                raise TypeError(f"duplicate param={p} in macro={self.name}")
+            s.add(p)
+
+    def call(self, namespace, args, kwargs):
+        return namespace._evaluator.start(
```
e-carlin, line 520: `_evaluator` should be public if accessed like this

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+        self.global_ns = PKDict()
+        self.local_ns = _Namespace(self)
+        self.global_fvars = PKDict()
+        self.local_fvars = PKDict()
+
+    def fconf_var(self, name):
+        try:
+            return self.global_fvars.pknested_get(name)
+        except KeyError:
+            raise KeyError(f"unknown macro param or fconf_var={name}")
```
e-carlin, line 300: Use raise from?

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+            PKDict(key=key, func=f.f_code.co_name, line=f.f_lineno),
+        )
+
+    def pop(self):
+        self.stack.pop()
+
+    def top(self):
+        if self.stack:
+            return self._str(self.stack[-1])
+        else:
```
e-carlin, line 493: No need for else and need `return "None"`

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+            "None"
+
+    def __str__(self):
+        return "/".join([k.key for k in self.xpath.stack])
```
e-carlin, line 497: `self.stack`

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+"constants" file, for example, to direct the flow of the subsequent files.
+
+:copyright: Copyright (c) 2022 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+
+"""
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdc, pkdlog, pkdp
+import contextlib
+import copy
```
e-carlin, line 137: unused

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+import inspect
+import pykern.pkrunpy
+import re
+import pykern.pkyaml
+import pykern.pkcollections
+import pykern.pkio
+
+
+#: matches word(word1, ...). Words have to begin with a lettera
```
e-carlin, line 146: typo letter

e-carlin (approved):
There are quite a few TODO, DOC, etc comments left in the code. I didn't comment on them but I think they can be cleaned up. Other than that and the some nitpicks I left it looks good.

pykern/fconf.py
```diff
@@ -0,0 +1,569 @@
...
+_RESERVED_PREFIX = "fconf_"
+
+_SELF = _RESERVED_PREFIX + "self"
+
+_NO_PARAM = object()
+
+_BUILTINS_EXT = "builtins"
+
+
+def parse_all(path):
```
mkeilman, line 170: I have noticed that if multiple files in the  directory import logging stuff, as in
```python
from pykern.pkdebug import pkdlog, pkdp
```
then the parser fails with a "duplicate macro" error. I haven't seen it fail with other duplicate imports yet. ???

robnagler: 7be8cce

garsuga (approved):

pykern/fconf.py
robnagler: Yes, and it's called by a dispatch so path is always passed.

pykern/fconf.py
robnagler: It's not public, because I don't want clients of fconf to use it this way. This would be called "protected" in Java, but it's not really. All the classes except Parser are private. _Namespace is tricky because all macros are operating in this namespace so they can access its attributes. `__getattr__` is only called if an attribute is not found. I think this is good practice (_attrs can be accessed within a module).

pykern/fconf.py
robnagler: See https://github.com/radiasoft/pykern/pull/224/files#r915385141

pykern/fconf.py
robnagler: 4766c22

pykern/fconf.py
robnagler: 4766c22

pykern/fconf.py
robnagler: 4766c22

pykern/fconf.py
robnagler: 4766c22

pykern/fconf.py
robnagler: 4766c22

robnagler:
@e-carlin writes:
> There are quite a few TODO, DOC, etc comments left in the code. I didn't comment on them but I think they can be cleaned up. Other than that and the some nitpicks I left it looks good.

It's still a work in progress. I'm going to leave them in until I finish rsconf migration.

PR #231, opened 2022-07-01
230 int keys

robnagler:
Merge after https://github.com/radiasoft/pykern/pull/224

- [x] Require test?
   - Yes
- [x] Security implications?
    - n/a
- [x] UI changes reviewed?
    - none
- [x] Backwards compatible?
    - yes, because aborted before
- [x] Require doc changes?
    - no
    - no


pykern/pkcollections.py
```diff
@@ -99,9 +101,68 @@ def pkdel(self, name, default=None):
...
+        def _type_err(key, base, merge):
+            return AssertionError(
+                f"key={key} type mismatch between (self) base={base} and to_merge={merge}"
+            )
+
+        if make_copy:
+            to_merge = copy.deepcopy(to_merge)
+
+        for k in list(to_merge.keys()):
+            t = to_merge[k]
```
gurhar1133, line 139: is `t` necessary?

pykern/pkcollections.py
```diff
@@ -112,7 +173,15 @@ def pknested_get(self, dotted_key):
         """
         d = self
         for k in dotted_key.split("."):
-            d = d[k]
+            try:
```
gurhar1133, line 176: why not type check d instead of nesting try blocks?

pykern/pkcollections.py
```diff
@@ -176,6 +242,54 @@ class PKDictNameError(NameError):
...
+        obj (object): what to convert
+
+    Returns:
+        object: converted object (may or may not be the same)
+    """
+    o = obj
+    if o is None:
+        return o
+    # Order matters so we don't convert bools to ints, since bools are ints.
+    for x in bool, int, float, str:
```
gurhar1133, line 277: What about this kind of thing instead of long list of if blocks?
```python
PKDict(
    bool=bool(o),
    int=int(o),
   ...
   collections.abc.Iterable=list(canonicalize(i) for i in iter(o))
)[type(o)]

gurhar1133 (commented):
Looks good 👍  Noticed potential typos and left a couple comments/questions.

pykern/pkcollections.py
robnagler: We don't know what `d` is, it's just something that responds to `__getitem__`. There is no coupling here except through TypeError which is the interface when `__getitem__` gets the wrong type of key. Then we try integer. If there's any particular error, we raise the original error so that it looks as it did before. This way we don't break abstractions.

pykern/pkcollections.py
robnagler: I think so, because it is used 10+ times, and `to_merge[k]` is a function call (under the covers) so this type of caching is important in recursive routine.

robnagler:
Changes submitted in in https://github.com/radiasoft/pykern/pull/239 and https://github.com/radiasoft/pykern/pull/237

Issue #205, opened 2022-06-17
RFC: pykern.quest


robnagler:
A Quest is "a long or arduous search for something." It is also a Pythonesque word so appropriate. The reason it is called Quest is that it is longer than a simple function call, since it must validate its inputs, e.g. an HTTP request, command line execution (pkcli), a Web Socket message, etc.

There are several classes in sirepo.quest:

* Spec is the decorator that describes the functions inputs and other prerequisites (e.g. permissions)
* Call is instantiated to make a call to a quest function
* Args is a PKDict subclass that holds the inputs
* Result is the return value or reply

Sirepo will subclass these, e.g. a sirepo.api.Base is a subclass of pykern.quest.Call, and used as follows:

```py
    @api_perm.require_user
    def api_deleteSimulation(self):
        req = self.parse_post(id=True)
        simulation_db.delete_simulation(req.type, req.id)
        return self.reply_ok()
```

Would turn into:

```py
    @sirepo.api.Spec('require_user', id='SimId')
    def api_deleteSimulation(self):
        simulation_db.delete_simulation(self.args.type, self.args.id)
        self.done(status='ok')
```

`self.result(status='ok')` would be shortened to `self.result_ok()`.

In the above example, sirepo.api.Spec always adds `type='SimType'` and applies the `@api_perm.require_user`. The value `auth_uid` is also in args. All values in args are validated, and collisions are managed safely, e.g. if a caller tries to supply auth_uid, it will be rejected, since that type of arg is special (the structures for this have to be defined clearly).

args will be parsed from whatever inputs make sense for the way the Call is made. If it is an API request and there is a URI with parameters, those values will be extracted, validated, and add to args. If there is a POST or PUT, those values will be validated and added to args. This is already being done to a degree with parse_post, but this formalizes all inputs.

The types of the inputs are found in pykern.pktype. For Sirepo sirepo.pktype will define its additional types. The type interface has yet to be defined, but this is part of the project.

e-carlin:
Sounds good. A few thoughts:

Cargill says "why" is an important question. I'll try to answer that on my own but I'd like your feedback to make sure we're on the same page. The api_perm part is more or less the same as we are doing now. The new part is moving parse_post (or the lack there of) into a defined and always included location. That's a good why. With all of the jupyter/oauth stuff it became clear to me that leaving calling parse_post up to the user creating the api is dangerous. Are there other whys you think are important?

[...]

PR #243, opened 2022-07-11
#205 quest pktype

mkeilman:
Here are some possible schema type definitions for discussion in a draft pr

### DIFF
--- a/pykern/pktype.py
+++ b/pykern/pktype.py
@@ -11,10 +11,10 @@ from pydantic import BaseModel
 import typing
 
 
-class Base():
+class Base:
     def validate(self, val):
         # require override?
-        #assert 0, 'subclass must override validate()'
+        # assert 0, 'subclass must override validate()'
         return val
 
 
@@ -24,12 +24,12 @@ class PKBoolean(Base):
         return val
 
 
-class PKChoices():
+class PKChoices:
     def __init__(self, choices):
         self.choices = frozenset(choices)
 
     def validate(self, val):
-        assert val in self.choices, ValueError(f'value={val} not in {self.choices}')
+        assert val in self.choices, ValueError(f"value={val} not in {self.choices}")
         return val
 
 
@@ -55,7 +55,9 @@ class PKRangedInt(PKInt):
 
     def validate(self, val):
         v = super().validate(val)
-        assert (self.min_val is None or v >= self.min_val) and (self.max_val is None or v <= self.max_val)
+        assert (self.min_val is None or v >= self.min_val) and (
+            self.max_val is None or v <= self.max_val
+        )
         return v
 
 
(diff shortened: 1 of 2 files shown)


pykern/pktype.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""Validated types
+
+:copyright: Copyright (c) 2022 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
```
robnagler, line 7: delete

pykern/pktype.py
```diff
@@ -0,0 +1,80 @@
...
+from pykern.pkcollections import PKDict
+from pykern import pkconfig
+
+
+class PKType():
```
robnagler, line 12: The types themselves, since they are capitalized, don't need to be PKType. We will never refer to them unqualified. Therefore, this type should be pktype.Base. 

The larger issue is that we instances of these types so that we aren't constantly referring to the schema to type check a result or whatever. This gets pretty tricky. You would use [`__new__` ](https://docs.python.org/3/reference/datamodel.html#object.__new__) for immutable types (int, bool, str, frozenset, etc.). The `__new__` would take as its argument, say, a string, and return an instance of cls. If this failed, it would raise ValueError or TypeError, similar to what int, bool, etc. do.

To convert these values to strings, I think we might want to consider a separate class that knows how to dispatch conversions to other classes. If this is the case, it might be better to not subclass the immutable types, and instead use them directly. 

Another alternative is multiple inheritance, but I like that a lot less.

IOW, these needs a bit more thought and/or discussion.

pykern/pktype.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""Validated types
+
+:copyright: Copyright (c) 2022 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
+from pykern.pkcollections import PKDict
+from pykern import pkconfig
```
robnagler, line 9: Don't need

pykern/pktype.py
robnagler: https://github.com/radiasoft/sirepo/issues/4167#issuecomment-1073254224 discusses pydantic. We should look at existing solutions to see if they do enough of what we want.

mkeilman: The `typing` module might even suffice; I'm looking at that now. I didn't know about it until browsing pydantic.

mkeilman:
I played around with pydantic which has the advantage of some built-in validation, see pktype_pydantic.py. Note that pydantic validates at initialization, not when values are set later, so
```python
r = RangedFloat(min_val=0, val=0.5)
r.val = -1.0
```
does not raise an exception. Maybe enough for our purposes though. Anyway, does pydantic look useful enough to pursue  further?

mkeilman:
Closing this draft pr

Issue #223, opened 2022-06-28
Add workflow for merged pull requests


robnagler:
[PR merge workflow from GitHub Docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges):

```yml
on:
  pull_request:
    types:
      - closed

jobs:
  if_merged:
    if: github.event.pull_request.merged == true
    runs-on: ubuntu-latest
    steps:
    - run: |
        echo The PR was merged
```

We don't know what happens here. Is there a checkout, and what is it checking out (main/master)?

Test on your own repo first to figure out what variables are set. Will radia_run ci-pull-request work out of the box?

rorour:
There is a checkout, but it doesn't check out any commits on our branches- I think we discussed this a bit when we were setting up github actions. [It checks out GITHUB_REF](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) which in this case has the contents of the merge (I just tested this, [see also](https://github.com/actions/checkout/issues/15#issuecomment-524093065)). 

So every time we update the pr (aka the 'synchronize' action), it tests on the potential merge, not just our feature branch. If we want to also run the tests when we merge the branch (I assume we want this in the case that master has been updated since the tests were run on the feature branch) we just need to add `types: [opened, reopened, synchronize, closed]` since the default is `types: [opened, reopened, synchronize]`. Is that what you had in mind?

robnagler:
I think the big one is closed with merged check. We always delete the branch on merge-close so I don't think we need much more than that. 

This is a distinct workflow from ci-pull-request. Perhaps they could be combined, but that may be tricky.

robnagler:
If you can make them a single workflow, great!

rorour:
Let me know if this PR is what you had in mind. 

> I assume we want this in the case that master has been updated since the tests were run on the feature branch

If this assumption was correct, is testing after merging the best way to handle this case? wouldn't it be safer to require merging master into the feature branch and run tests before merging?

robnagler:
Tests before merging would be great, but there's still a race condition between the review and the merge. There may be no merge conflicts, but someone else might have merged other PRs between the time of the last merge and the merge of the current PR. I don't think there's a good way to do this but making sure this happens on all merged PRs as they are merged to main.

PR #246, opened 2022-07-12
fix #234: GitHub action api verification

### LINKED ISSUE #234: GitHub Action verification that HEAD's SHA has passed all actions
There needs to be a way of checking github to see if HEAD has passed ci-pull-request workflow. Maybe there's a way to search actions via an API. The goal is to ask the question before a release: has HEAD passed CI?

rorour:
fix #234: GitHub Action verification that HEAD's SHA has passed all actions

Here is something functional. Looking forward to feedback.

- Vars owner and branch are just temporary while I'm developing. I'm testing with rorour/pykern because it has workflow runs on master.
- In order to get the commit sha matching HEAD from the API, I changed the post-merge action to run on master/main instead of the merged branch. 
- Using requests because the github3 python library doesn't support actions (sigmavirus24/github3.py/issues/1006)
- Issues to consider:
   - How to handle repos that use main instead of master?  I tried to make 'branch' a parameter with default value but that didn't seem to work with pkcli, I got `pykern github: error: unrecognized arguments: main`
   - How to get the HEAD sha that we're checking against the api? my approach assumes that we have a local copy of the repo, and checks out and pulls master.
   - How to authenticate with github - for now I'm using os.environ['GITHUB_PAT']


pykern/pkcli/github.py
```diff
@@ -326,6 +326,43 @@ def restore(git_txz):
...
+    r = requests.get(
+        f"{_GITHUB_API}/repos/{owner}/{repo}/actions/runs?per_page=1&page=1&branch={branch}",
+        headers={
+            "Accept": "application/vnd.github+json",
+            "Authorization": "token " + os.environ["GITHUB_PAT"],
+        },
+    )
+    assert r.json().get("workflow_runs"), f"no workflow runs for branch {branch}"
+    api_head_sha = r.json().get("workflow_runs")[0].get("head_sha")
+    _run_on_repo("checkout", branch)
```
robnagler, line 354: The github3 documentation doesn't seem to be up to date on the web. I know this, because I read the code.
I found the [latest_sha](https://github.com/sigmavirus24/github3.py/blob/1df2461e830092413778d5f3ff2a8785821321cb/src/github3/repos/branch.py). Then I saw [check_runs](https://github.com/sigmavirus24/github3.py/blob/1df2461e830092413778d5f3ff2a8785821321cb/src/github3/repos/commit.py#L44) which does have a [conclusion](https://github.com/sigmavirus24/github3.py/blob/1df2461e830092413778d5f3ff2a8785821321cb/src/github3/checks.py#L208). 

I wrote this code in github.py which I think can do what you need which also reuses some of the code in github.py:

```py
def t(repo, branch=None):
    r = _repo_arg(repo)

    def _branch(r, name, reraise=True):
        try:
            return r.branch(name=name)
        except github3.exceptions.NotFoundError:
            if reraise:
                raise

    b = _branch(r, branch) if branch else _branch(r, "master", False) or _branch(r, "main")
    pkdp(b.latest_sha())
    pkdp([c.conclusion for c in b.commit.check_runs()])
```

Here's how it works:

```sh
$ pykern github t radiasoft/sirepo
../pykern/pykern/pkcli/github.py:131:t 8bf6d7c2fdf28c9979075d305d1988c4484a786a
../pykern/pykern/pkcli/github.py:132:t []
$ pykern github t radiasoft/sirepo --branch mantid
../pykern/pykern/pkcli/github.py:131:t 4d3d1322613c96a8b04d09f0443b2f1ef2550556
../pykern/pykern/pkcli/github.py:132:t ['success', 'success']
```

.github/workflows/ci-pull-request.yml
```diff
@@ -1,14 +1,16 @@
...
+      - main
+      - master
+  push:
     branches:
       - main
       - master
 
 jobs:
   ci-pull-request:
-    if: ${{ github.event.action != 'closed' || github.event.pull_request.merged == true }}
```
robnagler, line 11: Does it have to run on all event types? Seems like we don't want a closed PR that isn't going to be merged, for example.

robnagler (changes requested):

.github/workflows/ci-pull-request.yml
robnagler: Ah, I see. Thanks.

pykern/pkcli/github.py
```diff
@@ -326,6 +326,43 @@ def restore(git_txz):
...
+        )
+
+    r = requests.get(
+        f"{_GITHUB_API}/repos/{owner}/{repo}/actions/runs?per_page=1&page=1&branch={branch}",
+        headers={
+            "Accept": "application/vnd.github+json",
+            "Authorization": "token " + os.environ["GITHUB_PAT"],
+        },
+    )
+    assert r.json().get("workflow_runs"), f"no workflow runs for branch {branch}"
```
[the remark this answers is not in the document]
rorour: Does this apply to all the asserts or just this one?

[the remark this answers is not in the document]
robnagler: That's the question for you: Assertions indicate the program is operating incorrectly. How do you tell the difference between an assertion and an error in external inputs?

pykern/pkcli/github.py
rorour: b09f39dd081e33c2fea46f0f81697bb23d7c3e7f

rorour:
@robnagler ready for review

pykern/pkcli/github.py
```diff
@@ -49,6 +50,44 @@ def backup():
...
+    s = b.commit.sha
+    _run_on_repo("checkout", b.name)
+    _run_on_repo("pull")
+    o = _run_on_repo("rev-parse", f"origin/{b.name}")
+    if not s == o:
+        raise pkcli.command_error(f"Most recent run on {b.name}={s} does not match {o}")
+    c = [c.conclusion for c in b.commit.check_runs()]
+    if not c:
+        raise pkcli.command_error("No workflow runs for commit")
+    if "success" not in c or any(x not in ("success", "skipped") for x in c):
```
robnagler, line 86: I think you need to verify that the most recent run is successful. I don't know really what the content of check_runs can be, but I'm guessing if you rerun a check (for transient errors, for example), then it will add more runs.

robnagler (changes requested):

pykern/pkcli/github.py
```diff
@@ -49,6 +50,44 @@ def backup():
     pkdlog("DONE")
 
 
+def ci_check(repo, branch=None):
+    def _branch(r, name, reraise=True):
+        try:
+            return r.branch(name=name)
+        except github3.exceptions.NotFoundError:
+            if reraise:
+                raise
```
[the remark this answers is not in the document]
rorour: 0be93f0d99229da03b7ee0293c13867f08425e1d

pykern/pkcli/github.py
```diff
@@ -49,6 +50,44 @@ def backup():
...
+    )
+    s = b.commit.sha
+    _run_on_repo("checkout", b.name)
+    _run_on_repo("pull")
+    o = _run_on_repo("rev-parse", f"origin/{b.name}")
+    if not s == o:
+        raise pkcli.command_error(f"Most recent run on {b.name}={s} does not match {o}")
+    c = [c.conclusion for c in b.commit.check_runs()]
+    if not c:
+        raise pkcli.command_error("No workflow runs for commit")
```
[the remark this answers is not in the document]
rorour: 0be93f0d99229da03b7ee0293c13867f08425e1d

pykern/pkcli/github.py
robnagler: What is it skipping? That is, are there two workflows and it is skipping one because it doesn't match the criteria? The reason I'm asking is the test is rather complicated. It would be easier to have: c[0] == "success" or something like that.

robnagler (commented):
I think this is ok, but I would like to resolve the conditional complexity.

pykern/pkcli/github.py
rorour: Yes, I was testing on a branch that had two jobs in the same workflow. If we'll only ever have one job per workflow then  c[0] == "success" makes sense. 
In checking this I realized that the change to the workflow file needs to be made in sirepo too, and I never changed it in sirepo when I fixed #223. I'll open a pr for that as well.

rorour: 92f8cb5ae798c6456ced913c26ada41747101e27

pykern/pkcli/github.py
```diff
@@ -49,6 +50,30 @@ def backup():
...
+    b = (
+        _branch(r, branch)
+        if branch
+        else _branch(r, "master", False) or _branch(r, "main")
+    )
+    s = b.commit.sha
+    c = [c.conclusion for c in b.commit.check_runs()]
+    if not c:
+        pkcli.command_error("No workflow runs for commit")
+    if not c[0] == "success":
```
robnagler, line 72: `c[0] != "success"`

pykern/pkcli/github.py
```diff
@@ -49,6 +50,30 @@ def backup():
...
+    r = _repo_arg(repo)
+    b = (
+        _branch(r, branch)
+        if branch
+        else _branch(r, "master", False) or _branch(r, "main")
+    )
+    s = b.commit.sha
+    c = [c.conclusion for c in b.commit.check_runs()]
+    if not c:
+        pkcli.command_error("No workflow runs for commit")
```
robnagler, line 71: Include branch name and sha. Watch for repetition when you do this to the next error message to.

robnagler (changes requested):

pykern/pkcli/github.py
rorour: What do you mean by watch for repetition? Should I just print branch name and sha once before all the error checking?

robnagler: First step is add the branch and sha to all messages. Then, see what patterns repeat. Then, simplify the repetition.

pykern/pkcli/github.py
rorour: 49014a8d17a76e7fb62bc12a07d7e337831e44f7

pykern/pkcli/github.py
rorour: pksubprocess_test keeps randomly failing on this pr. Do you know what might be causing it?

```
=================================== FAILURES ===================================
_________________________ test_check_call_with_signals _________________________
  File "/home/runner/work/pykern/pykern/tests/pksubprocess_test.py", line 65, in test_check_call_with_signals
    pksubprocess.check_call_with_signals(cmd, output=o, msg=msg)
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/_pytest/python_api.py", line 966, in __exit__
    fail(self.message)
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/_pytest/outcomes.py", line 196, in fail
    raise Failed(msg=reason, pytrace=pytrace)
Failed: DID NOT RAISE <class 'RuntimeError'>
============================== 1 failed in 10.15s ==============================
error: FAILED=1 passed=37
```

robnagler: No, I don't. How would you instrument pksubrocess_test to debug this? How would you debug this more easily?

rorour: I think the failing test is unrelated to the changes in this pr, so I can make an issue for it. This pr is ready for re-review. ea3c408a5e353e43282c3a154677ec387b309992

robnagler (approved):
That's what I was looking for.

PR #258, opened 2022-08-05
Fix #252 ExceptToFile catches exceptions properly

### LINKED ISSUE #252: Better error when pkexcept_to_file used for both deviance and conformance
If there is an error in xlsx_test in the conformance case, pkexcept is written. It ends with a:

```txt
FileNotFoundError: [Errno 2] No such file or directory: '/Users/nagler/src/radiasoft/pykern/tests/xlsx_work/2/xl/sharedStrings.xml'
```

But this is not what's happened in the conformance case. case_dirs should look for pkexcept and compare it first. If it isn't there, then it should output the result. May also want the stack trace to another file.

robnagler:
- [x] Require test?
   - yes
- [x] Security implications?
    - none
- [x] UI changes reviewed?
    - Someone should look at output to see if makes sense
- [x] Backwards compatible?
    - no, but fixed all uses
- [x] Require doc changes?
    - yes
    - no


pykern/pkunit.py
```diff
@@ -50,6 +57,42 @@ class PKFail(AssertionError):
...
+    Used for deviance testing with `case_dirs`.
+
+    If there is an exception, writes that to the file. Otherwise, writes "None"
+
+    If there is an exception, will write `PKSTACK_PATH`. Otherwise, no
+    file exists. Used for diagnostics only.
+
+    Usage::
+        for d in case_dirs():
+        with ExceptToFile():
```
gurhar1133, line 72: indent the with block?

gurhar1133 (approved):
Looks good 👍

pykern/pkunit.py
robnagler: 4ac8732

PR #271, opened 2022-08-23
Fix #268: ignore lines

### LINKED ISSUE #268: case_dirs/file_eq needs to take ingore lines as input
case_dirs and file_eq need to be able to take a list of lines to be ignored for file comparison

gurhar1133:
ignore_lines list keyword arg to case_dirs or file_eq will now ignore all lines that match regex in list
test added


pykern/pkunit.py
```diff
@@ -129,6 +130,9 @@ def case_dirs(group_prefix=""):
     of these expect files is copmared to the corresponding `work_dir`
     actual file using `file_eq`.
 
+    Ignore_lines is a list of posix style regex. See `Posix style regex syntax
+    <https://www.gnu.org/software/findutils/manual/html_node/find_html/posix_002dbasic-regular-expression-syntax.html>_`
```
robnagler, line 134: Remove this, and specify Args: (below) and say:

ignore_prefix (iterable): `POSIX standard regular expressions <https://www.gnu.org/software/findutils/manual/html_node/find_html/posix_002dbasic-regular-expression-syntax.html>`_ to be passed to ``diff``

The syntax for rST is 

```rst
`phrase <link>`_
```

robnagler (changes requested):
There are conflicts you'll need to fix as well.

mkeilman (commented):
Just a few comments, though I could not test because of merge conflicts.

pykern/pkunit.py
```diff
@@ -509,18 +513,36 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
             line_terminator="\r\n",
         )
 
+    def _gen_ignore_lines(self):
+        r = []
+        for l in self._ignore_lines:
+            if l:
+                r.append("-I")
```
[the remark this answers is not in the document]
robnagler: I have to say that comprehension always bothered me. Here's a shorter one I just discovered: `filter(None, self._ignore_lines)`. When the filter is `None` returns the truthy values only. `itertools.filterfalse` does the opposite, which is why I tried the first case.

pykern/pkunit.py
```diff
@@ -509,18 +513,36 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
...
+                r.append(l)
```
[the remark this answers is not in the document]
robnagler: Again, I was curious. I don't have a lot of experience with itertools:

`list(itertools.chain.from_iterable((itertools.zip_longest((), ('a', 'b', 'c'), fillvalue='-I'))))`

I don't think this is worth it.

tests/pkunit_test.py
```diff
@@ -178,6 +178,23 @@ def test_xlsx_to_csv_conversion():
...
 
 
+def test_ignore_lines():
+    from pykern import pkunit
+
+    d = pkunit.data_dir()
+    pkunit.file_eq(
+        expect_path=d.join("ignore_lines.in"),
+        actual_path=d.join("ignore_lines.out"),
+        ignore_lines=[r"[0-9]\+xyz"],
```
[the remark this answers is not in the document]
robnagler: Ha ha. I told him to do it the way it is now. I think having the data in files for a "file_eq" test makes more sense.

gurhar1133:
@robnagler conflicts resolved

pykern/pkunit.py
```diff
@@ -510,20 +514,44 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
             line_terminator="\r\n",
         )
 
+    def _gen_ignore_lines(self):
+        r = []
+        if self._ignore_lines is None:
+            return r
+        for l in (l for l in [l for l in self._ignore_lines if l]):
```
robnagler, line 521: There are too many `for` loops. This is a better encapsulation, which both shortens the code and separates out a cohesive chunk:

```py
def _cmd():
    if self.is_bytes:
        r = ["cmp"]
    else:
        r = ["diff"]
        for l in self._ignore_lines or ():
            r.extend(("-I", l))
    return r + [str(self._expect_path), str(self._actual_path)]
```

Move `_cmd` inside `_compare`. You don't need to pass self.

`_gen_ignore_lines` is non-descriptive. 

`ignore_lines` could be [], (), None, False, etc. We don't care, it's falsey, and do nothing. 

Use tuples unless you expect the object is going to be modified. 

The elements of ignore_lines must be strings. If someone wants to pass "" as an element, pass that as a value to -I.

pykern/pkunit.py
```diff
@@ -510,20 +514,44 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
...
+            r.extend(["-I", l])
+        return r
+
     def _compare(self):
         if self._expect == self._actual:
             return
-        c = f"{'cmp' if self.is_bytes else 'diff'}  '{self._expect_path}' '{self._actual_path}' 2>&1"
-        with os.popen(c) as f:
+
```
robnagler, line 528: no blank line.

pykern/pkunit.py
```diff
@@ -510,20 +514,44 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
...
+        c = ["cmp"] if self.is_bytes else ["diff", *self._gen_ignore_lines()]
+        c.extend(
+            [
+                f"{self._expect_path}",
+                f"{self._actual_path}",
+            ]
+        )
+        p = subprocess.run(
+            c,
```
robnagler, line 537: `p=subprocess.run(_cmd(), stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True)`

mkeilman (approved):
lgtm!

pykern/pkunit.py
```diff
@@ -510,20 +514,44 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
...
             pkfail(
-                "{}",
                 f"""expect != actual:
-    {c}
-    {f.read()}
-    {self._message()}
-    """,
+'{"' '".join(c)}'
+{p.stdout}
+{self._comparison_message(p.returncode)}""",
```
robnagler, line 547: This should be a nested function, too. Describe what it is, and again cohesive modularization. There's an indentation issue in _message (those lines should be left justfied), and combine in a single nested function:

```py
def _failed_msg(process):
        r = "'" + "' '".join(process.args) + f"'\n" + p.stdout + "\n"
        if not (return_code == 1 or self.is_bytes):
            return r + "diff command failed\n"
        if self._expect_is_jinja:
            return r + f"""
Implementation restriction: expect is a jinja template which has been processed to
produce the diff. A simple copy of actual to expect is not possible. You will need to update
the expect jinja template={self._expect_path} manually.
"""
        return r + f"""
to update test data:
    cp '{self._actual_path}' '{self._expect_path}'
"""
```

robnagler (changes requested):

gurhar1133:
@robnagler made changes

pykern/pkunit.py
```diff
@@ -510,18 +514,54 @@ def _actual_xlsx_to_csv(self, actual_xlsx, sheet):
             line_terminator="\r\n",
         )
 
+    def _gen_ignore_lines(self):
```
robnagler, line 517: not used, delete. same for _message().

robnagler (changes requested):

pykern/pkunit.py
gurhar1133: oops, sorry i missed those

gurhar1133:
@robnagler how bout now?

robnagler (approved):

PR #278, opened 2022-09-26
Fix #205 pykern.quest basic module

### LINKED ISSUE #205: RFC: pykern.quest

robnagler:
- [x] Require test?
   - no, it's a skeleton module
- [x] Security implications?
    - n/a
- [x] UI changes reviewed?
    - n/a
- [x] Backwards compatible?
    - yes
- [x] Require doc changes?
    - n/a
    - n/a

### DIFF
--- a/pykern/quest.py
+++ b/pykern/quest.py
@@ -9,16 +9,20 @@ from pykern.pkdebug import pkdc, pkdlog, pkdp
 
 
 class Args:
+    # qargs
     pass
 
 
-class API:
+class Callable(PKDict):
+    # qcall
     pass
 
 
 class Result:
+    # qres
     pass
 
 
 class Spec:
+    # qspec
     pass
--- a/pykern/quest.py
+++ b/pykern/quest.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+"""API wrapper
+
+:copyright: Copyright (c) 2022 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdc, pkdlog, pkdp
+
+
+class Args:
+    pass
+
+
+class API:
+    pass
+
+
+class Result:
+    pass
+
+
+class Spec:
+    pass
--- a/pykern/quest.py
+++ b/pykern/quest.py
@@ -13,7 +13,7 @@ class Args:
     pass
 
 
-class Callable(PKDict):
+class API(PKDict):
     # qcall
     pass
 


gurhar1133 (approved):
Don't completely understand what this is a skeleton for, but I don't see any issues with it

pykern/quest.py
```diff
@@ -0,0 +1,28 @@
...
+from pykern.pkdebug import pkdc, pkdlog, pkdp
+
+
+class Args:
+    # qargs
+    pass
+
+
+class API(PKDict):
+    # qcall
```
e-carlin, line 17: This is the only one not like the others. Name the class Call?

robnagler: The instance is a call. The class is an API. I know if it is "confusing", but it's something we can get used to. The semantics are important. You cannot say API().api_beginRequest (as is done on master), because you have to qcall.call_api("beginRequest") to call that method. However, all we have is the concept of "self", which is the qcall, which allows the API during a  call to call useful methods, e.g. self.reply_ok. These utilities are not APIs. I thought it was prudent to not have to say self.qcall.reply_ok. Perhaps that is a mistake.

robnagler:
> Don't completely understand what this is a skeleton for, but I don't see any issues with it

See https://github.com/radiasoft/sirepo/pull/4880 which has sirepo.quest.

PR #281, opened 2022-09-27
Fix #280: removed todo, require ext arg

### LINKED ISSUE #280: pkcli.github glob for checking for previous clone is unnecessary
The code is 
```python
    def _prev_backup(self, base, ext=_TXZ):
        # POSIT: timestamp Backup
        b = pkio.sorted_glob(f"../*/{base}{ext}")
        return b[-1] if b else []
```

In the case of zgoubi git we have 
base = radiasoft-zgoubi.git
ext = "*"

That matches both radiasoft-zgoubi.git and radiasoft-zgoubi.github.io.git.

So we cp in radiasoft-zgoubi.github.io.git
https://github.com/radiasoft/pykern/blob/0f782f55bb0eee315f9b675f10a3e9bc0567e074/pykern/pkcli/github.py#L447

And the chdir doesn't work because radiasoft-zgoubi.git doesn't exist
https://github.com/radiasoft/pykern/blob/0f782f55bb0eee315f9b675f10a3e9bc0567e074/pykern/pkcli/github.py#L448

e-carlin:
In the event the glob matches multiple values. For example, radiasoft-zgoubi.git and radiasoft-zgoubi.github.io.git. Then use the first result which will be the shorter and more "specific" result.

- [ ] Require test?
   - N
- [ ] Security implications?
    - N
- [ ] UI changes reviewed?
    - N
- [ ] Backwards compatible?
    - I hope... hard to test
- [ ] Require doc changes?
    - N
- [ ] Add [DesignHints](https://github.com/radiasoft/pykern/wiki/DesignHints)?
    - N

@robnagler this seems rather challenging to test and I'm concerned about edge cases I'm not thinking of. Any thoughts?

### DIFF
--- a/pykern/pkcli/github.py
+++ b/pykern/pkcli/github.py
@@ -409,7 +409,7 @@ class _Backup(_GitHub):
         pkdc("updating from {}", backup)
         _shell(["tar", "xJf", str(backup)])
 
-    def _prev_backup(self, base, ext=_TXZ):
+    def _prev_backup(self, base, ext):
         # POSIT: timestamp Backup
         b = pkio.sorted_glob(f"../*/{base}{ext}")
         return b[-1] if b else []
@@ -440,11 +440,7 @@ class _Backup(_GitHub):
                     ),
                 )
                 return
-            # TODO(robnagler) remove backward compatibility after 2022-09-01
-            if prev.ext == _TXZ:
-                self._extract_backup(prev)
-            else:
-                _shell(("cp", "--archive", "--link", str(prev), "./"))
+            _shell(("cp", "--archive", "--link", str(prev), "./"))
             with pkio.save_chdir(base):
                 _shell(["git", "remote", "update"])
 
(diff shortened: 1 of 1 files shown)


e-carlin:
On closer thought. Maybe the "*" glob can be removed entirely. What case is that covering?

e-carlin:
I updated to remove the glob. I don't think there was any use for it.

pykern/pkcli/github.py
```diff
@@ -427,7 +427,7 @@ def _repo(self, repo):
 
         def _clone(suffix):
             base = bd + suffix
-            prev = self._prev_backup(base, ext="*")
+            prev = self._prev_backup(base, ext="")
```
robnagler, line 430: isn't it needed for wiki?

e-carlin: I don't believe so. In the wiki case we'll look for 'radiasoft-<repo>.wiki.git' which will work. Rigth?

PR #283, opened 2022-09-30
Fix #251: pkcli.message

### LINKED ISSUE #251: add pkcli.message for output that is not an error but a warning or info
This would be similar to command_error without the exception. The message would go to stderr.

gurhar1133:
pkcli.message writes to stderr without raising
includes a test


pykern/pkcli/__init__.py
```diff
@@ -72,6 +72,15 @@ def command_error(fmt, *args, **kwargs):
     raise CommandError(fmt.format(*args, **kwargs))
 
 
+def message(fmt, *args, **kwargs):
```
robnagler, line 75: `command_info` is a better name, since it relates to `command_error`.

tests/pkcli_data/m1.in/x.py
```diff
@@ -0,0 +1,5 @@
+import pykern.pkcli
+import sys
+
+pykern.pkcli.message("this is a message")
```
robnagler, line 4: This is not the intended usage. It should only be called from a pkcli. Better to structure it that way so people don't use it as a example (tests as documentation).

tests/pkcli_data/m1.in/x.py
```diff
@@ -0,0 +1,5 @@
+import pykern.pkcli
```
robnagler, line 1: Not sure I understand the naming of this. We had p1, p2, and now m1.in. Perhaps just 1.in? The m1 makes me ask "what does m stand for"?

tests/pkcli_test.py
```diff
@@ -64,6 +64,17 @@ def test_main3():
...
 
 
+def test_message():
+    from pykern import pksubprocess
+    from pykern.pkunit import case_dirs
+
+    for d in case_dirs("m1"):
+        pksubprocess.check_call_with_signals(
+            ["python", d.join("x.py")],
+            output="res.txt",
```
robnagler, line 74: res.txt maybe better x.out?

robnagler (changes requested):

gurhar1133:
made changes @robnagler

robnagler (approved):

PR #288, opened 2022-10-02
Fix #287 added skip_past to pkcli.test

### LINKED ISSUE #287: 20221002 pkcli.test start_path=file
Start at a particular file so you can test incrementally.

robnagler:
- [x] Require test?
   - added
- [x] Security implications?
    - n/a
- [x] UI changes reviewed?
    - please do
- [x] Backwards compatible?
    - yes
- [x] Require doc changes?
    - done
    - no

### DIFF
--- a/pykern/pkcli/test.py
+++ b/pykern/pkcli/test.py
@@ -28,8 +28,9 @@ def default_command(*args):
     An argument which is ``case=<pattern>``, is passed to pytest
     as ``-k <pattern>``.
 
-    ``skip_to=<first_test>`` causes collection to ignore all files
-    before ``<first_test>``` (may be partial match).
+    ``skip_past=<last_to_ignore>`` causes collection to ignore all
+    files up to and including ``<last_to_ignore>``` (may be partial
+    match of the test file).
 
     Writes failures to ``<base>_test.log``
 
@@ -37,6 +38,7 @@ def default_command(*args):
         args (str): test dirs, files, options
     Returns:
         str: passed=N if all passed, else raises `pkcli.Error`
+
     """
     from pykern import pkconfig
     from pykern import pksubprocess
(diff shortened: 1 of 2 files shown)


pykern/pkcli/test.py
```diff
@@ -94,33 +99,47 @@ def default_command(*args):
...
-    i = re.compile(r"(?:_work|_data)/")
     res = []
+
+    def _file(path):
+        nonlocal skip_past
+        if skip_past:
+            if skip_past in path:
+                skip_past = None
+            return
+        res.append(path)
```
gurhar1133, line 130: I might be missing something here, but it seems like this is saying, if the skip path is in the path, dont append it, otherwise append it. But how are we assuring all files up to skip path are being skipped as well?

gurhar1133 (approved):
Check my comments, if my concerns aren't relevant, then I approve

pykern/pkcli/test.py
```diff
@@ -28,12 +28,17 @@ def default_command(*args):
...
+    files up to and including ``<last_to_ignore>``` (may be partial
+    match of the test file).
+
     Writes failures to ``<base>_test.log``
 
     Args:
         args (str): test dirs, files, options
     Returns:
         str: passed=N if all passed, else raises `pkcli.Error`
+
```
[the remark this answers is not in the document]
robnagler: Yes, https://github.com/radiasoft/pykern/blob/9551826282bb90d1a4d73219af726637da75665c/pykern/pkexample.py#L73

pykern/pkcli/test.py
robnagler: Walk through the logic again including line 126. 128 only is hit after all the files have been skipped before skip_past matches.

gurhar1133: Oh, gotcha

PR #308, opened 2022-10-20
fix #307: handle print in comment, better error message

### LINKED ISSUE #307: 20221115 Change "error" to "check_prints error"
```bash
$ echo "# i am a comment with print('foo')" > x.py
$ pykern ci check_prints
error: x.py:1 # i am a comment with print('foo')
```

I think the _PRINT_OK regex should be changed
from: `^\s*#\s*(?:pkdp|print)\(`
to: `^\s*#.*(?:pkdp|print)\(` (There can be any characters after `#` not just whitespace)

The command error should say where the error is coming from (what function). When it is run in `pykern ci run` the reader has no clue where the error is coming from (see https://github.com/radiasoft/rsconf/actions/runs/3292642290/jobs/5428245294 for example).

e-carlin:


pykern/pkcli/ci.py
```diff
@@ -43,6 +43,9 @@ def check_prints():
     from pykern import pkio
     from pykern import pkcli
 
+    def _error(msg):
+        pkcli.command_error("check_prints: {}", msg)
```
e-carlin, line 47: I doubt you'll like this but lmk what should be here. Some message that "this occured in check_prints" would be helfpul otherwise one is left scratching their head when CI fails like https://github.com/radiasoft/rsconf/actions/runs/3292642290/jobs/5428245294#step:3:370

pykern/pkcli/ci.py
```diff
@@ -24,7 +24,7 @@
     + r"|^run/"
 )
 _PRINT = re.compile(r"(?:\s|^)(?:pkdp|print)\(")
-_PRINT_OK = re.compile(r"^\s*#\s*(?:pkdp|print)\(")
+_PRINT_OK = re.compile(r"^\s*#.*(?:pkdp|print)\(")
```
robnagler, line 27: This is insufficient. A docstring with a print() does not work, and there were several that had to be fixed to pass check_prints.

The point really is that we are trying to be very strict about print statements. The case you pointed out didn't require a `print()`. If you want to avoid the check, you can always put a space between the `print` and `(`. However, I don't recommend that.

robnagler (changes requested):

robnagler (approved):

PR #317, opened 2022-10-31
fix #305: check for newline at eof

### LINKED ISSUE #305: 20221031 pkcli.ci check for newline at eof
Try for js, tsx, py, yml, etc. files.

rorour:
check_eof_newline and check_prints are very similar, but I'm not sure if it makes sense to make an abstraction for them.


pykern/pkcli/ci.py
```diff
@@ -35,23 +66,17 @@ def check_prints():
     wiki for an explanation of why this check is necessary.
 
     If you really need a print, use `pykern.pkconst.builtin_print`.
-
-    Args:
-        exclude (regex): path pattern to exclude from check
-    ..
```
rorour, line 41: does `..` mean something? I saw you added it @robnagler f15946e86deddae7f27402ad41e73dc06cd19ba2

robnagler: No, that's a typo. I don't think there are any args to check_prints so please remove.

pykern/pkcli/ci.py
```diff
@@ -8,25 +8,56 @@
 :copyright: Copyright (c) 2022 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
-from pykern.pkdebug import pkdp, pkdlog
```
robnagler, line 11: This line states in, because it makes debugging convenient.

pykern/pkcli/ci.py
```diff
@@ -8,25 +8,56 @@
...
+    """Recursively check repo for files missing newline at end of file.
+
+    Files matching _CHECK_EOF_NEWLINE_FILE_EXTS and not matching
+    _CHECK_EOF_NEWLINE_EXCLUDE_FILES will be checked.
+    """
+
+    def _error(msg):
+        pkcli.command_error("check_eof_newline: {}", msg)
+
+    res = []
```
robnagler, line 44: How could this code be shared with check_prints?

tests/pkcli_data/command_info.in/example_pkcli_module.py
```diff
@@ -4,4 +4,4 @@
 def _example_pkcli_function():
     pykern.pkcli.command_info("this is a message")
 
-_example_pkcli_function()
\ No newline at end of file
+_example_pkcli_function()
```
robnagler, line 7: Need a test for check_eof_newline().

pykern/pkcli/ci.py
```diff
@@ -8,25 +8,56 @@
...
+from pykern import pkio
 from pykern import pksetup
+from pykern import pkunit
 import re
 
-
-_FILE_TYPE = re.compile(r".py$")
-_EXCLUDE_FILES = re.compile(
+_CHECK_EOF_NEWLINE_EXCLUDE_FILES = re.compile(
+    r"/ext/|node_modules/|^run/|tests/[\w\d/]+_work/|^venv/"
```
robnagler, line 18: This is very specific to sirepo. Also, we do want to skip test data like check_prints_exclude_files does.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: e987d4cecc20f01b5a8d29ed4f44274e61aaeb0a

pykern/pkcli/ci.py
rorour: What's the harm in having these here for any repo? Do you foresee us running this on another repo in the future where we would want to check for newlines in some of these directories?

robnagler: We are going to run this on all repos, eventually. The scientists want to implement CI. It's running on rslinac now.

/ext/ is the only one that's problematic. If it was more qualified, e.g. /static/js/ext/, I'd be less concerned.

pykern/pkcli/ci.py
```diff
@@ -73,6 +67,51 @@ def run():
...
+
+    def _check_prints(file_path, res):
+        for j, l in enumerate(pkio.read_text(file_path).split("\n"), start=1):
+            if re.search(_PRINT, l) and not re.search(_PRINT_OK, l):
+                res.append(f"{file_path}:{j} {l}")
+
+    def _error(m):
+        pkcli.command_error("{}: {}", case, m)
+
+    if case == "check_eof_newline":
```
gurhar1133, line 89: Another option (but maybe @robnagler can determine if a better option) would be to do dictionary dispatch eg.
```python
d = PKDict(
    check_eof_newline=PKDict(...),
    check_prints=PKDict(...),
)[case]
```
Would have be modified to `raise ValueError(...)` if case is not valid (unless KeyError would be ok)

gurhar1133 (commented):
One comment but looks good! 👍

pykern/pkcli/ci.py
robnagler: KeyError is fine, because _check_files is an internal routine.

pykern/pkcli/ci.py
```diff
@@ -73,6 +67,51 @@ def run():
...
+        d = PKDict(
```
robnagler, line 90: This PKDict could be a constant instead of having to map the names, and it could be a pkdict of pkdicts:

```py
PKDict(check_prints=PKDict(exclude_files=re.compile(
    r"/ext/|node_modules/|^run/|tests/|^venv/"
), 
etc.
```

pykern/pkcli/ci.py
```diff
@@ -8,59 +8,53 @@
...
 _PRINT_OK = re.compile(r"^\s*#\s*(?:pkdp|print)\(")
 
 
+def check_eof_newline():
+    """Recursively check repo for files missing newline at end of file.
+
+    Files matching _CHECK_EOF_NEWLINE_FILE_EXTS and not matching
+    _CHECK_EOF_NEWLINE_EXCLUDE_FILES will be checked.
+    """
+    _check_files("check_eof_newline")
```
robnagler, line 42: You could simplify the code and remove dispatch by calling _check_files this way:

```py
def check_eof_newline():
    def _file(lines):
        return None if lines[-1] == "" else ""

    return _check_files("check_eof_newline", _file)
```

None means no error. Have _check_files prefix the file_path to the result and append to res so you don't have to pass either. Both routines need lines so share that code.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: 951542b61ce1c38b913cce136408231f0994e067

tests/pkcli_data/command_info.in/example_pkcli_module.py
rorour: a44f8b2fb1e292d1140765110bbb79627445f1f2

pykern/pkcli/ci.py
rorour: 728c88a7bcadafc9b100216121fe8d7f8ac81c72

pykern/pkcli/ci.py
rorour: f2008e1c66a39ead6e0c01f83feddc1c1b0d258d

pykern/pkcli/ci.py
rorour: f2008e1c66a39ead6e0c01f83feddc1c1b0d258d

pykern/pkcli/ci.py
rorour: 16e58b893588afe93755e152fe7397ee18ff7eaf

gurhar1133 (approved):
Looks good

rorour:
should probably merge https://github.com/radiasoft/sirepo/pull/5065 first

pykern/pkcli/ci.py
```diff
@@ -8,59 +8,52 @@
...
-    + r"|^run/"
-)
 _PRINT = re.compile(r"(?:\s|^)(?:pkdp|print)\(")
 _PRINT_OK = re.compile(r"^\s*#\s*(?:pkdp|print)\(")
 
 
+def check_eof_newline():
+    """Recursively check repo for files missing newline at end of file.
+
+    Files matching _CHECK_EOF_NEWLINE_FILE_EXTS and not matching
```
robnagler, line 26: comment refers to non-existent vars. Better to describe what it does in plain English.

pykern/pkcli/ci.py
```diff
@@ -73,6 +66,50 @@ def run():
...
+    r = []
+    p = pkio.py_path()
+    n = 0
+    for f in pkio.walk_tree(p, d.include_files):
+        f = p.bestrelpath(f)
+        if re.search(d.exclude_files, f):
+            continue
+        n += 1
+        c = check_file(pkio.read_text(f).split("\n"))
+        if c is not None:
```
robnagler, line 108: I see that None was not the best choice. Rather, return an empty list if there are no errors. Then the code would be simpler. You could just say `for l in check_file(...)` and not need the temporary.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: 7ad30ac

tests/pkcli/ci_test.py
```diff
@@ -4,25 +4,24 @@
 :copyright: Copyright (c) 2022 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
+from pykern import pkunit
```
robnagler, line 7: In pytests, it is important to not import anything globally, except pytest. Move to the test methods.

tests/pkcli/ci_test.py
```diff
@@ -4,25 +4,24 @@
...
-import pytest
 
 
-def test_run():
-    from pykern.pkcli import ci
-    from pykern import pkio
-    from pykern import pkunit
-
-    for d in pkunit.case_dirs("run"):
+def test_check_eof_newline():
```
robnagler, line 12: You have three cases which are the same. I think you could use the case_dir name to determine which method to call with a simple regex and renaming the directories. For example, rename run1.* to run-1.* and then you could even use str.split to get the name of the function.

pykern/pkcli/ci.py
```diff
@@ -73,6 +64,48 @@ def run():
...
     check_prints()
     fmt.diff(pkio.py_path())
     test.default_command()
+
+
+def _check_files(case, check_file):
+    def _error(m):
+        pkcli.command_error("{}: {}", case, m)
+
+    d = PKDict(
```
robnagler, line 77: This dictionary could be a global constant like `_PRINT`, and it would simplify the code here.

pykern/pkcli/ci.py
```diff
@@ -73,6 +64,48 @@ def run():
...
+        check_eof_newline=PKDict(
+            exclude_files=re.compile(
+                r"/static/js/ext/|node_modules/|^run/|tests/|^venv/"
```
robnagler, line 80: A little documentation here might help. It's ok to mention that static/js/ext is where Sirepo stores 3rd party library files. I think we may want to ignore "^tests/" not just "tests/". Scientists use the word "tests" to mean conditions to test. Narrowing here would be the comparison is useful. Similarly, `^node_modules/` is probably a good idea.

pykern/pkcli/ci.py
```diff
@@ -8,59 +8,50 @@
...
-    + r"|^run/"
-)
 _PRINT = re.compile(r"(?:\s|^)(?:pkdp|print)\(")
 _PRINT_OK = re.compile(r"^\s*#\s*(?:pkdp|print)\(")
 
 
+def check_eof_newline():
+    """Recursively check repo for files missing newline at end of file.
+
+    Files specified in _check_files() in exclude/include_files will be checked.
```
robnagler, line 26: While "by reference" documentation is easy to maintain, it's not useful. Please denormalize the description. It doesn't have to be literal, e.g. it might say something like ignore directories that match run, test data/work, package_data, etc. Try to imagine reading this as an outsider on pykern.rtfd.io.

pykern/pkcli/ci.py
```diff
@@ -8,59 +8,50 @@
...
+    def _c(lines):
+        return [] if lines[-1] == "" else [""]
+
+    _check_files("check_eof_newline", _c)
+
+
 def check_prints():
     """Recursively check repo for (naked) print and pkdp calls.
 
+    Files specified in _check_files() in exclude/include_files will be checked.
```
robnagler, line 38: Ditto above.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: Since this dict is only used in this function, I don't understand why it's better to make it global. It makes this part of the code look cleaner but seems to me like it's actually more complicated by spreading this logic to different parts of the file.

robnagler: Good point. I think the thing I see is a long function which immediately makes it more difficult for me to read. If I see a constant at the top of the file, it's an abstraction (similar to a separate function) that I know is constant so in general, I ignore it unless I need to know the details. This then simplifies the function so the business logic is clearer. Feel free to ask others for opinions.

tests/pkcli/ci_test.py
rorour: a9e80a0beda32d4163def682c4cffe18637e6076

tests/pkcli/ci_test.py
rorour: is this what you had in mind? a9e80a0beda32d4163def682c4cffe18637e6076

pykern/pkcli/ci.py
rorour: a9e80a0beda32d4163def682c4cffe18637e6076

pykern/pkcli/ci.py
rorour: We need to include react/node_modules so I think it makes sense to keep `node_modules/` as it is.
a9e80a0beda32d4163def682c4cffe18637e6076

pykern/pkcli/ci.py
rorour: a9e80a0beda32d4163def682c4cffe18637e6076

pykern/pkcli/ci.py
rorour: a9e80a0beda32d4163def682c4cffe18637e6076

tests/pkcli/ci_test.py
robnagler: Almost. _test could be test_all:

```py
def test_all():
    from pykern import pkunit
    from pykern.pkcli import ci

    for d in pkunit.case_dirs(case_dir):
        with pkunit.ExceptToFile():
            getattr(ci, d.basename.split("-")[0])()
```

pykern/pkcli/ci.py
```diff
@@ -73,6 +85,30 @@ def run():
...
     fmt.diff(pkio.py_path())
     test.default_command()
+
+
+def _check_files(case, check_file):
+    def _error(m):
+        pkcli.command_error("{}: {}", case, m)
+
+    d = _CHECK_FILES[case]
+
```
robnagler, line 99: Remove space

robnagler (approved):

pykern/pkcli/ci.py
rorour: [dd3d04c](https://github.com/radiasoft/pykern/pull/317/commits/dd3d04c2cec070bd6c7a42300b4a8f2f38a58fde)

tests/pkcli/ci_test.py
robnagler: It's fully declarative. The directories describe the tests and the test script is just the interpreter, which will be just one function, does not need to be changed when new kinds of tests are added.

rorour: The code you gave for test_all still uses case_dir. I don't see how this works without adding another loop around it like `for case_dir in ("check_eof_newline", "check_prints", "run")`. Could you explain that?

robnagler: Sorry I meant to remove case_dir as an argument to case_dirs. It's not necessary. If you read my original comment, it explains how this works. Try out the suggested file renaming.

rorour: 92d66abb3d1b0843ba74b3d09480199b674dd55b

PR #322, opened 2022-11-20
Fix #321 added pkinspect.package_module_names

### LINKED ISSUE #321: pkinspect.package_modules: list of modules in a package

robnagler:


pykern/pkinspect.py
```diff
@@ -70,6 +72,9 @@ def __str__(self):
             return "<no file>:0:<no func>"
 
 
+clsmembers = inspect.getmembers(sys.modules[__name__], inspect.isclass)
```
e-carlin, line 75: This doesn't look like it is used anywhere. Also, if keeping it, should it be a const?

e-carlin (approved):

pykern/pkinspect.py
robnagler: d007b03 was a note to self...

PR #325, opened 2022-11-30
Fix #311: file_eq ndiff

### LINKED ISSUE #311: file_eq should be able to ndiff files
file_eq should be able to ndiff files that have a .ndiff extension, just like with .jinja and .json

gurhar1133:
file_eq runs ndiff on .ndiff extension files.


gurhar1133:
@robnagler  I pytest.skip if ndiff is not available. On github test that means the ndiff test will be skipped. Haven't opened the 
accompanying radiasoft/container-python3 issue to make ndiff available. Not sure if that should be completed before this is merged.

pykern/pkunit.py
```diff
@@ -555,6 +560,30 @@ def _failed_msg(process):
...
+            )
+
+        def _ndiff_files(expect_path, actual_path, epsilon):
+            _ndiff_config(epsilon, w)
+            p = subprocess.run(
+                (
+                    "ndiff",
+                    actual_path,
+                    expect_path,
+                    w.join(_NDIFF_CONF_FILE),
```
robnagler, line 575: This is what I call implicit coupling. Return the file path from _ndiff_config so that you don't have the implicit coupling of work_dir and NDIFF_CONF_FILE. Just call it in place (no temp var) and remove `_NDIFF_CONF_FILE` constant, because it is unnecessary.

pykern/pkunit.py
```diff
@@ -555,6 +560,30 @@ def _failed_msg(process):
...
+                ),
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+            )
+            d = pkcompat.from_bytes(p.stderr)
```
robnagler, line 580: Test the process status on `p` if ndiff does return a status (like `diff` does). Testing for a specific string isn't as robust especially if ndiff crashes for a different reason.

gurhar1133: So ndiff wont error if it finds diff (no fail fast just writes to file). Should I do something like
```python
if p.returncode != 0:
    pkfail("ndiff failed: {}", p.stderr)
```
just in case ndiff fails for some other reason and we want to know more?

robnagler: The failure cases outnumber the success cases so it is more robust then to test for the success case. This makes the code more robust. What does it output on success?

gurhar1133: on success it will just output how many lines have been diffed eg.
```
3 lines have been diffed
```

robnagler: So you would test for exactly `^\d+ lines have been diffed\n*$`. The goal is to know what success is without getting collisions on possible failures.

gurhar1133: Both failure and success will actually say 
```
n lines have been diffed
```
and it is failures that result in 
```
n diffs detected
```

So in that case would it make more sense to test p.stderr for `n diffs detected` ?

robnagler (approved):

PR #326, opened 2022-12-03
fix #313: ci check for modules with main programs

### LINKED ISSUE #313: 20221130 add ci check for modules with main programs
madx_device_server is an example of this. Only "main" should be in `<project>`_console.py.

rorour:


robnagler:
@rorour ci is failing. I'll wait till it passes to review.

robnagler (approved):

PR #330, opened 2022-12-16
fix #329: ci defaults to only look in <pkg> and tests

### LINKED ISSUE #329: 20221213 ci check_prints should only look in <pkg> and tests directories
https://github.com/radiasoft/rshellweg/pull/141 is failing because prints are found, but they were there all along. I wonder if something changed.

I think we may want to change the way check_prints and check_main is started. It should only run on the tests and `<pkg>`, e.g. sirepo, directories. I'm wondering if we don't want the same rule for fmt as well. Thoughts?

rorour:


pykern/pkcli/ci.py
```diff
@@ -124,8 +124,14 @@ def _error(m):
...
         f = p.bestrelpath(f)
-        if re.search(d.exclude_files, f):
+        if re.search(d.exclude_files, f) or (
+            github_ci_repo
+            and not re.search(
+                re.compile(github_ci_repo + r"/|tests/|README|setup\.py$"), f
+            )
+        ):
             continue
+
```
robnagler, line 134: Unnecessary whitespace. If you feel the need for whitespace, probably time to make a separate function

pykern/pkcli/ci.py
```diff
@@ -44,7 +44,7 @@
 _PRINT_OK = re.compile(r"^\s*#\s*(?:pkdp|print)\(")
 
 
-def check_eof_newline():
+def check_eof_newline(github_ci_repo=None):
```
robnagler, line 47: In general, this type of parameter is better solved with config. It would be passed in python-ci downloader. However, I don't think we need to do it this way. Rather, we know this is a python ci. If it is executed in the repo root, the file setup.py will be there, for example. This can be a sentinel for specific action. In this case, what we want to limit is the top level recursions. In the walk_tree on line 125, just list the directories that are supposed to be included, e.g. package and tests. This will require an outer loop for `p` for this special case. This could be solved by creating a function that yields (generator) results for the outerloop. In the top level case, it would yield two directories. The other cases would yield a single directory.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: fmt added that

pykern/pkcli/ci.py
rorour: Can we meet about this on monday? I don't understand your description, and looking back I think I misinterpreted [your comment](https://github.com/radiasoft/pykern/issues/329#issuecomment-1352277598).

rorour: figured it out, although I didn't see the advantage of using a generator for this case. Will send an email monday about the confusion above. 928b447ead65201a9275ac9d42f7f202d7eb5bf4

rorour:
sorry, just realized i forgot to include setup.py in the ci check. adding that now.

robnagler (changes requested):
I think I didn't think of the larger context when we had our discussion. The problem is the conversion to py.path which is necessary. `_check_files` is only called in the pwd without args. This means bestrelpath will always work.

pykern/pkcli/ci.py
```diff
@@ -120,15 +120,27 @@ def _error(m):
...
-    for f in pkio.walk_tree(p, d.include_files):
-        f = p.bestrelpath(f)
-        if re.search(d.exclude_files, f):
-            continue
-        n += 1
-        for l in check_file(pkio.read_text(f).split("\n")):
-            r.append(f"{f}{l}")
+
+    def _paths():
+        if pkio.py_path().join("setup.py").isfile() and (
```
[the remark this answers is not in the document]
rorour: 488f8f30f8c5708beec57efc767533c435d2994b

pykern/pkcli/ci.py
```diff
@@ -120,15 +120,27 @@ def _error(m):
...
+            pkio.py_path().join("README.rst").isfile()
+            or pkio.py_path().join("README.md").isfile()
+        ):
+            return (
+                pkio.py_path().join(pkio.py_path().basename),
+                pkio.py_path().join("tests"),
+            ), str
+        return (pkio.py_path(),), pkio.py_path().bestrelpath
+
+    l, g = _paths()
```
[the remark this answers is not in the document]
rorour: 488f8f30f8c5708beec57efc767533c435d2994b

pykern/pkcli/ci.py
```diff
@@ -120,15 +120,28 @@ def _error(m):
...
-        for l in check_file(pkio.read_text(f).split("\n")):
-            r.append(f"{f}{l}")
+
+    def _paths():
+        if pkio.py_path().join("setup.py").isfile() and (
+            pkio.py_path().join("README.rst").isfile()
+            or pkio.py_path().join("README.md").isfile()
+        ):
+            return (
+                pkio.py_path().join(pkio.py_path().basename),
```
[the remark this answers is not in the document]
rorour: 488f8f30f8c5708beec57efc767533c435d2994b

pykern/pkcli/ci.py
```diff
@@ -19,11 +19,11 @@
 #: static/js/ext is where Sirepo stores 3rd party library files
 _CHECK_FILES = PKDict(
     check_eof_newline=PKDict(
-        exclude_files=re.compile(r"/static/js/ext/|node_modules/|^run/|^tests/|^venv/"),
+        exclude_files=re.compile(r"/static/js/ext/|node_modules/|run/|tests/|venv/"),
```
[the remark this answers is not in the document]
rorour: 488f8f30f8c5708beec57efc767533c435d2994b

pykern/pkcli/ci.py
```diff
@@ -120,15 +120,27 @@ def _error(m):
...
+            return (
+                pkio.py_path().basename,
+                "tests",
+                "setup.py",
+            )
+        return (pkio.py_path(),)
+
+    for p in _paths():
+        for f in pkio.walk_tree(p, d.include_files):
+            f = pkio.py_path().bestrelpath(f)
```
robnagler, line 138: pkio.py_path() is a loop invariant. Need to pull that out of the loop, because getting cwd is expensive relative to the operations.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: 2793e544f3899eefb18ae9d4c895cdffdc1483c8

PR #332, opened 2022-12-23
fix #329: fmt.diff checks same paths as ci._check_files

### LINKED ISSUE #329: 20221213 ci check_prints should only look in <pkg> and tests directories

rorour:
The problem I mentioned in today's meeting was in fact [a bug that was fixed](https://github.com/psf/black/issues/2598) in version 22.12.0, and we only required >=22.0 (I was running 22.6.0.)

This update is enforcing new requirements, hence the change to pykern/pkcli/github.py and pykern/xlsx.py.

rshellweg passes ci format checks on my machine with these changes, but running `pykern fmt run .` will change files that `pykern ci run` does not diff. (Running `pykern fmt run .` in rshellweg will format those example files that were failing `pykern ci run`). Do we want `pykern fmt run .` to only run on `<pkg>`, `tests`, `setup.py `when run from the repo root?

### DIFF
--- a/pykern/pkcli/ci.py
+++ b/pykern/pkcli/ci.py
@@ -110,7 +110,7 @@ def run():
     check_eof_newline()
     check_main()
     check_prints()
-    fmt.diff(_paths(pkio.py_path()))
+    fmt.diff(*_paths(pkio.py_path()))
     test.default_command()
 
 
--- a/pykern/pkcli/ci.py
+++ b/pykern/pkcli/ci.py
@@ -110,7 +110,7 @@ def run():
     check_eof_newline()
     check_main()
     check_prints()
-    fmt.diff(pkio.py_path())
+    fmt.diff(_paths(pkio.py_path()))
     test.default_command()
 
 
@@ -121,17 +121,6 @@ def _check_files(case, check_file):
     d = _CHECK_FILES[case]
     r = []
     n = 0
-
-    def _paths(cwd):
-        if pkio.py_path("setup.py").isfile() and (
-            pkio.py_path("README.rst").isfile() or pkio.py_path("README.md").isfile()
-        ):
-            return (
-                cwd.basename,
-                "tests",
-                "setup.py",
-            )
-        return (pkio.py_path(),)
 
     c = pkio.py_path()
     for p in _paths(c):
@@ -147,3 +136,15 @@ def _check_files(case, check_file):
         _error("no files found")
     if r:
         _error("\n".join(r))
+
+
+def _paths(cwd):
+    if pkio.py_path("setup.py").isfile() and (
+        pkio.py_path("README.rst").isfile() or pkio.py_path("README.md").isfile()
+    ):
+        return (
+            cwd.basename,
+            "tests",
+            "setup.py",
+        )
+    return (pkio.py_path(),)
--- a/pykern/pkcli/ci.py
+++ b/pykern/pkcli/ci.py
@@ -139,10 +139,11 @@ def _check_files(case, check_file):
 
 
 def _paths(cwd):
-    if pkio.py_path("setup.py").isfile() and (
-        pkio.py_path("README.rst").isfile() or pkio.py_path("README.md").isfile()
+    if cwd.join("setup.py").isfile() and (
+        cwd.join("README.rst").isfile() or cwd.join("README.md").isfile()
     ):
         return (
+            # POSIT: repo name
             cwd.basename,
             "tests",
             "setup.py",
--- a/pykern/pkcli/ci.py
+++ b/pykern/pkcli/ci.py
@@ -148,4 +148,4 @@ def _paths(cwd):
             "tests",
             "setup.py",
         )
-    return (pkio.py_path(),)
+    return (cwd,)
--- a/pykern/pkcli/fmt.py
+++ b/pykern/pkcli/fmt.py
@@ -37,7 +37,9 @@ def check(*paths):
         _black(paths, "--check")
     except RuntimeError as e:
         if str(e) == "error exit(1)":
-            pykern.pkcli.command_error("paths={} need to be formatted", paths)
+            pykern.pkcli.command_error(
+                "paths={} need to be formatted", tuple(str(p) for p in paths)
+            )
         raise
 
 
--- a/pykern/pkcli/fmt.py
+++ b/pykern/pkcli/fmt.py
@@ -13,7 +13,7 @@ def run(*paths):
     """Run black formatter on `path`
 
     Args:
-        path (object): string or py.path to file or directory
+        *paths (strs or py.paths): strings or py.paths to file or directory
     """
     _black(paths)
 
@@ -22,22 +22,22 @@ def diff(*paths):
     """Run diff on file comparing formatted vs. current file state
 
     Args:
-        path (object): string or py.path to file or directory
+        *paths (strs or py.paths): strings or py.paths to file or directory
     """
     _black(paths, "--diff", "--check", "--no-color")
 
 
-def check(path):
+def check(*paths):
     """Returns True if there would be diff else return False
 
     Args:
-        path (object): string or py.path to file or directory
+        *paths (strs or py.paths): strings or py.paths to file or directory
     """
     try:
-        _black(path, "--check")
+        _black(paths, "--check")
     except RuntimeError as e:
         if str(e) == "error exit(1)":
-            pykern.pkcli.command_error("path={} needs to be formatted", path)
+            pykern.pkcli.command_error("paths={} need to be formatted", paths)
         raise
 
 
@@ -45,6 +45,7 @@ def _black(paths, *args):
     """Helper function invokes black with options
 
     Args:
+         *paths (strs or py.paths): strings or py.paths to file or directory
          *args (strs): options to be passed to black
     """
     from pykern import pkunit
--- a/pykern/pkcli/fmt.py
+++ b/pykern/pkcli/fmt.py
@@ -9,28 +9,22 @@ import py
 import pykern.pksubprocess
 
 
-def run(path):
+def run(*paths):
     """Run black formatter on `path`
 
     Args:
         path (object): string or py.path to file or directory
     """
-    _black(path)
+    _black(paths)
 
 
-def diff(path_or_paths):
-    """Run diff on file comparing formated vs. current file state
+def diff(*paths):
+    """Run diff on file comparing formatted vs. current file state
 
     Args:
         path (object): string or py.path to file or directory
     """
-    for p in (
-        (path_or_paths,)
-        if isinstance(path_or_paths, py._path.local.LocalPath)
-        or isinstance(path_or_paths, str)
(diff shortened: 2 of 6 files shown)


robnagler:
>  Do we want pykern fmt run . to only run on <pkg>, tests, setup.py when run from the repo root?

Correct. We don't want to enforce standards on non-standard directories.

pykern/pkcli/ci.py
```diff
@@ -110,7 +110,7 @@ def run():
     check_eof_newline()
     check_main()
     check_prints()
-    fmt.diff(pkio.py_path())
+    fmt.diff(_paths(pkio.py_path()))
```
robnagler, line 113: `*_paths()` is the way to expand a list/tuple.

pykern/pkcli/fmt.py
```diff
@@ -17,13 +18,19 @@ def run(path):
     _black(path)
 
 
-def diff(path):
+def diff(path_or_paths):
```
robnagler, line 21: `def diff(*paths)` (see below). This allows calling with multiple paths from the command line.

pykern/pkcli/fmt.py
```diff
@@ -17,13 +18,19 @@ def run(path):
...
     """Run diff on file comparing formated vs. current file state
 
     Args:
         path (object): string or py.path to file or directory
     """
-    _black(path, "--diff", "--check", "--no-color")
+    for p in (
+        (path_or_paths,)
+        if isinstance(path_or_paths, py._path.local.LocalPath)
```
robnagler, line 29: Don't reach inside`py`. `_path` is private. The way to do this is [done in pkconst](https://github.com/radiasoft/pykern/blob/10d95c24369e847fc12085bdce9239fe7d12e3c4/pykern/pkconst.py#L26) and exported. However, it's better to change the definition of `fmt` (above).

pykern/pkcli/fmt.py
```diff
@@ -17,13 +18,19 @@ def run(path):
...
+        or isinstance(path_or_paths, str)
+        else path_or_paths
+    ):
+        _black(p, "--diff", "--check", "--no-color")
```
robnagler, line 33: Pass `paths` to black without expansion. This is a private interface and doesn't interact with the command line. Probably better to pass the args as a single tuple, too:

```py
_black(paths, ("--diff", "--check", "--no-color"))
```

robnagler (changes requested):
Thanks for checking for bugs in black.

pykern/pkcli/fmt.py
rorour: As far as I can tell, black doesn't accept multiple paths. If we want to only call black once we would need to use the `--include` flag with a regex based on the paths. Otherwise we need a loop somewhere. Should `_black` loop through paths?

robnagler:
```sh
echo 'ok=1' > x.py
echo 'invalid python' > y.py
black --no-color x.py y.py
^[[31merror: cannot format y.py: Cannot parse: 1:8: invalid python^[[0m
^[[1mreformatted x.py^[[0m

^[[1mOh no! 💥 💔 💥^[[0m
^[[34m^[[1m1 file ^[[0m^[[1mreformatted^[[0m, ^[[31m1 file failed to reformat^[[0m.
```

rorour: ah sorry. I think it was pksubprocess.check_call_with_signals causing the issue because it needs them separated

rorour:
> > Do we want pykern fmt run . to only run on , tests, setup.py when run from the repo root?
> 
> Correct. We don't want to enforce standards on non-standard directories.

How should fmt run handle this? It seems like if fmt run is called from the command line we need to check each path that it receives and run it through ci._paths()?

robnagler:
> It seems like if fmt run is called from the command line we need to check each path that it receives and run it through ci._paths()?

`fmt` is a utility that abstracts our formatter. It might eventually know how to run JS formatter, for example. It does not need to know about repos except that it does need to know about exclusions since it delegates tree walking to the formatter, which is typical of formatters.

`ci` abstracts continuous integration, which includes formatting and other standards (check prints). `ci` is what enforces standards on a repo so it sets standards by calling `fmt` only where we set standards (tests, setup.py, etc.). Therefore, fmt doesn't need to be changed except to accept a list of files instead of a single directory.

rorour:
Then should we add a new command to ci to replace our widespread use of `pykern fmt run .`?

robnagler:
> Then should we add a new command to ci to replace our widespread use of pykern fmt run .?

Not sure what you mean. Developers can still make sure the whole repo is formatted. 

I'm not sure it is very widespread. For example, I never use `pykern fmt run .`, because I run [`pykern fmt run <file-saved>`](https://github.com/biviosoftware/home-env/blob/e324322d59d7db7f0ecbeb3029221000cdf5b7ad/emacs.d/b-python.el#L115). I'm pretty sure Evan and Paul use the same command. I suspect VS code and PyCharm could do the same. This way there's no need to run `fmt` manually, which means there won't be `ci` failures for people forgetting to run `pykern fmt run .` manually. 

Perhaps I'm misunderstanding.

rorour:
> > Do we want pykern fmt run . to only run on , tests, setup.py when run from the repo root?
> 
> Correct. We don't want to enforce standards on non-standard directories.

This comment made me think you wanted some kind of change to the way fmt run currently works

robnagler:
> This comment made me think you wanted some kind of change to the way fmt run currently works

Ah, no. fmt enforces standards, of course.

pykern/pkcli/ci.py
rorour: 81558f6e653269c07a7286f77898de84b03955fc

pykern/pkcli/fmt.py
rorour: 81558f6e653269c07a7286f77898de84b03955fc

pykern/pkcli/fmt.py
rorour: 81558f6e653269c07a7286f77898de84b03955fc

pykern/pkcli/fmt.py
rorour: 81558f6e653269c07a7286f77898de84b03955fc

pykern/pkcli/ci.py
```diff
@@ -147,3 +136,15 @@ def _paths(cwd):
         _error("no files found")
     if r:
         _error("\n".join(r))
+
+
+def _paths(cwd):
+    if pkio.py_path("setup.py").isfile() and (
```
robnagler, line 142: pkio.py_path("setup.py") is better written as cwd.join("setup.py"). All the py.path() can be cwd.

pykern/pkcli/ci.py
```diff
@@ -147,3 +136,15 @@ def _paths(cwd):
...
+        pkio.py_path("README.rst").isfile() or pkio.py_path("README.md").isfile()
+    ):
+        return (
+            cwd.basename,
```
robnagler, line 146: This line needs to be documented. This is a the repo name, but that's a posit based on how we setup repos.

robnagler (changes requested):

pykern/pkcli/ci.py
rorour: 3b6bc4ce71f665b83fc9f9cf3024c5434a631d09

pykern/pkcli/ci.py
rorour: 3b6bc4ce71f665b83fc9f9cf3024c5434a631d09

setup.py
```diff
@@ -13,7 +13,7 @@
     author_email="pip@pykern.org",
     install_requires=[
         "argh>=0.26",
-        "black>=22",
+        "black==22.12",
```
rorour, line 16: @robnagler black's updates are not backwards compatible for our purposes. Run on sirepo, the difference between 22.12 and the latest 23.1a1 is:

```
diff --git a/sirepo/template/flash.py b/sirepo/template/flash.py
index 048b361df..73bcd0286 100644
--- a/sirepo/template/flash.py
+++ b/sirepo/template/flash.py
@@ -418,7 +418,10 @@ def stateless_compute_update_lib_file(data):
             _SIM_DATA.flash_app_lib_basename(data.args.archiveLibId),
         ).copy(p)
     else:
-        simulation_db.simulation_dir(SIM_TYPE, data.args.simulationId,).join(
+        simulation_db.simulation_dir(
+            SIM_TYPE,
+            data.args.simulationId,
+        ).join(
             _SIM_DATA.flash_app_archive_basename()
         ).rename(p)
     return PKDict(

```

robnagler: So the fixed version will not force a rewrite of the repo?

tests/pkcli/fmt_data/check1.out/pkexcept
```diff
@@ -1 +1 @@
-path=check1 needs to be formatted
+paths=(local('check1'),) need to be formatted
```
robnagler, line 1: This error message needs to be fixed. It should say "('check',)" not "(local('check'),)".

pykern/pkcli/ci.py
```diff
@@ -147,3 +136,16 @@ def _paths(cwd):
...
+    if cwd.join("setup.py").isfile() and (
+        cwd.join("README.rst").isfile() or cwd.join("README.md").isfile()
+    ):
+        return (
+            # POSIT: repo name
+            cwd.basename,
+            "tests",
+            "setup.py",
+        )
+    return (pkio.py_path(),)
```
robnagler, line 151: cwd

robnagler (changes requested):

setup.py
rorour: [Just these changes.](https://github.com/radiasoft/sirepo/pull/5295) mostly removing blank lines in function and class definitions

robnagler: What I was confused about was which changes were caused by which version. 23.1a1 will break more changes, but we can plan when that happens to avoid broken PR checks. For now let's fix the version to `black~=22.0`.

pykern/pkcli/ci.py
rorour: 55a883d8ded98dee7e8d67ebd2ef0656c9bd2a71

tests/pkcli/fmt_data/check1.out/pkexcept
rorour: 55a883d8ded98dee7e8d67ebd2ef0656c9bd2a71

setup.py
rorour: c3b091879b8a8cd266d4d1fe691be99e1d88166e

robnagler (approved):
Great job sticking with the updates! Thanks.

PR #334, opened 2023-01-02
Fix #333: cp prompt for ndiff

### LINKED ISSUE #333: ndiff needs a cp prompt
When ndiff finds diffs we need the cp prompt to copy actual to expect if actual is correct

gurhar1133:
_update_message is shared for ndiff and regular diff failures for when we want to cp actual to expect


robnagler (approved):
You need to fix the fmt, but ok to merge when works.

rorour (commented):
Looks good to me

PR #342, opened 2023-03-13
Fix #341: relative diff as default and abs diff option

### LINKED ISSUE #341: ndiff file_eq should use relative difference by default
Right now, ndiff file_eq uses absolute difference. This is a problem for comparing very large numbers. The default should be to use relative difference for comparisons and absolute difference should be an optional setting.

gurhar1133:
Now relative difference is the default comparison mode for file_eq with ndiff. For some instances it is important to be able to use absolute difference (some cases where there are a lot of values less than 1). So absolute difference comparison is still an option.

This feature was needed because this [test](https://github.com/radiasoft/rslaser/blob/a1c50e8f4f7738d6019664e2ba0bdbdadf6549b5/tests/pulse_test.py#L108-L130) has very large outputs and doesn't make sense to use absolute diff.


gurhar1133:
@robnagler tests passing. ready for review

pykern/pkunit.py
```diff
@@ -551,18 +551,21 @@ def _failed_msg(process):
                 )
             return r + self._update_message
 
-        def _ndiff_config(epsilon, work_d):
+        def _ndiff_config(epsilon, abs_epsilon, work_d):
             return pykern.pkio.write_text(
-                work_d.join("ndiff_conf.txt"), f"* * abs={float(epsilon)}"
+                work_d.join("ndiff_conf.txt"),
+                f"* * abs={float(abs_epsilon)}"
```
robnagler, line 557: I think epsilon should be rel_epsilon. There now needs to be a test to be sure that one or the other is passed. I comment on this below.

pykern/pkunit.py
```diff
@@ -669,6 +677,7 @@ def _validate_args(self, expect_path, *args, **kwargs):
         self._expect_is_jinja = self._expect_path.ext == ".jinja"
         self._is_ndiff = self._expect_path.ext == ".ndiff"
         self._ndiff_epsilon = kwargs.get("ndiff_epsilon", 1e-13)
```
robnagler, line 679: When changing the meaning of a parameter, you should maybe make a new parameter so call it ndiff_rel_epsilon. Then, in the code, you throw an error if someone passes ndiff_epsilon with a message about what's changed. You can deprecate, but since it's little used, it's probably not important.

This is a little used interface, but it's the principle that's important here. 

If is_ndiff, then rel or abs should be passed, but not both. You can choose an appropriate default (rel, 1e-13), but it should be after checking the parameters aren't conflicting.

pykern/pkunit.py
```diff
@@ -572,7 +575,12 @@ def _ndiff_files(expect_path, actual_path, epsilon):
                 pkfail("diffs detected: {} {}", d, self._update_message)
 
         if self._is_ndiff:
-            _ndiff_files(self._expect_path, self._actual_path, self._ndiff_epsilon)
+            _ndiff_files(
+                self._expect_path,
+                self._actual_path,
+                self._ndiff_epsilon,
```
robnagler, line 581: I think _ndiff_files should refer to self for the epsilon.

pykern/pkunit.py
```diff
@@ -551,18 +551,21 @@ def _failed_msg(process):
...
 
-        def _ndiff_files(expect_path, actual_path, epsilon):
+        def _ndiff_files(expect_path, actual_path, epsilon, abs_epsilon):
             p = subprocess.run(
                 (
                     "ndiff",
                     actual_path,
                     expect_path,
-                    _ndiff_config(epsilon, w),
+                    _ndiff_config(epsilon, abs_epsilon, w),
```
robnagler, line 568: use the values on self. Don't need to pass through.

tests/pkunit_test.py
```diff
@@ -78,6 +78,16 @@ def test_file_eq_ndiff():
         actual_path=d.join("actual_conformance.ndiff"),
         ndiff_epsilon=1e-9,
     )
+    pkunit.file_eq(
```
robnagler, line 81: test the deviance case of passing both rel and abs.  Test the case of passing ndiff_epsilon causing an assertion. 

This particular case names the diff files "rel" but the parameter is "abs". Don't use 1e-13, use something that's not the default.

tests/pkunit_test.py
```diff
@@ -78,6 +78,16 @@ def test_file_eq_ndiff():
...
+        expect_path=d.join("expect_rel.ndiff"),
+        actual_path=d.join("actual_rel.ndiff"),
+        abs_ndiff_epsilon=1e-13,
+    )
+    with pkunit.pkexcept("diffs detected:"):
```
robnagler, line 86: This is documented as "rel", but it's implicit. Make it the expect_default.ndiff, and then test the case of passing a rel epsilon. Rename the "conformance_2" to match the other names: rel, default, rel2, whatever.

robnagler (changes requested):

pykern/pkunit.py
gurhar1133: What do you think about passing just epsilon and a kwarg use_absolute for if someone wants to use absolute difference? Then there is only one epsilon.

robnagler: `ndiff_epsilon` and `ndiff_epsilon_is_abs`.

gurhar1133:
@robnagler lmk if more changes needed. I want to get this merged so I can fix some rslaser stuff

robnagler (approved):

PR #344, opened 2023-03-13
Fix #343 pkconfig._clean_environ prefers upper case values

### LINKED ISSUE #343: pkconfig flatten values should not complain about duplicate env values
flatten values should take the all upper case version of any value that's a duplicate.

robnagler:


gurhar1133 (approved):
Looks good to me. And thanks for fixing that xlsx test

PR #353, opened 2023-04-04
Fix #352 pkdp return value example doc

### LINKED ISSUE #352: Add example with pkdp return value to pkdebug docstring
This feature is mentioned in passing, but a concrete example would be helpful.

mkeilman:
I added an example but I'm not sure whether it's a good demo of this particular usage; suggestions welcome. Feel free to ignore the bonus wordsmithery...


pykern/pkdebug.py
```diff
@@ -54,6 +53,11 @@
     You can match any text in the line output with a regular expression, which
     is case insensitive.
 
+    If you omit args and kwargs, `pkdp` returns its argument, of use in
+    constructions such as::
```
e-carlin, line 57: Maybe instead of "of use in construction" you could say "of use when you want to inspect a value without changing the flow of the code"? Not sure my wording is any better...

You example is good.

FWIW another one I commonly use is 
```python
def foo():
...
  return pkdp(res)
```

e-carlin (approved):

pykern/pkdebug.py
```diff
@@ -40,10 +40,9 @@
...
 
-    If you do nothing, the first print statement would come
-    out always. The second wouldn't come out unless you
-    specified a ""control" via the environment variable
-    ``$PYKERN_PKDEBUG_CONTROL``:
+    The first print statement always appears.
+    The second does not unless you specify a "control" via the
+    environment variable ``$PYKERN_PKDEBUG_CONTROL``:
 
         PYKERN_PKDEBUG_CONTROL=my_mod python my_prog.py
```
robnagler, line 47: There is a shortcut for this value:

`        pkdebug=my_mod python my_prog.py`

Document both.

pykern/pkdebug.py
```diff
@@ -40,20 +40,34 @@
...
 
+    or equivalently::
```
robnagler, line 49: "or with the shortname `pkdebug`"

robnagler:
@mkeilman can this be merged?

mkeilman:
> @mkeilman can this be merged?

If the wording is ok, yes...I'll get it synced with master now

PR #360, opened 2023-04-13
Fix #359 add and use pkconfig.in_dev_mode

### LINKED ISSUE #359: pkconfig.in_dev_mode() to replace most channel_in("dev") uses
We have used `channel_in("dev")` to mean both "executing in a development environment" and "enable development features". These are two different modes, and they need to be controlled differently. 

`$PKERN_PKCONFIG_DEV_MODE` config will control `pkconfig.in_dev_mode()`. It will default to be true if the channel is "dev" so there won't be any change in behavior.

For example, in development, you want to enable [debug mode in Tornado in pkasyncio](https://github.com/radiasoft/pykern/blob/5cac3be0e2b66e332e5dcdba413281c264303ddf/pykern/pkasyncio.py#L90):

```py
        debug=(pkconfig.channel_in("dev"), bool, "enable debugging for asyncio"),
```

However, in a dev execution environment that is simulating deployment, we would not want this on, and this default should be controlled by `in_dev_mode`.

robnagler:
- fixed some pkconfig tests to use pkunit

### DIFF
--- a/pykern/pkconfig.py
+++ b/pykern/pkconfig.py
@@ -78,7 +78,7 @@ from pykern import pkinspect
 #: Python version independent value of string instance check
 STRING_TYPES = pkconst.STRING_TYPES
 
-#: Environment variable holding channel (defaults to 'dev')
+#: Environment variable holding channel (defaults to "dev")
 CHANNEL_ATTR = "pykern_pkconfig_channel"
 
 _KEY_PATTERN = "^[A-Z][A-Z0-9_]*[A-Z0-9]$"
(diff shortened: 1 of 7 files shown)


pykern/pkconfig.py
```diff
@@ -668,10 +679,13 @@ def _coalesce_values():
...
     _raw_values = values
     _parsed_values = dict(((_Key([k]), v) for k, v in env.items()))
-    cfg = init(
+    a = PKDict(
         _caller_module=sys.modules[__name__],
         channel=Required(str, "which (stage) function returns config"),
     )
+    cfg = init(**a)
+    a.dev_mode = (channel_in("dev"), bool, "controls development features")
+    cfg = init(**a)
```
gurhar1133, line 688: dont understand why a.dev_mode is being set in between two calls to `init()`

gurhar1133 (approved):
Looks good. Left one comment

pykern/pkconfig.py
robnagler: The first call sets `cfg.channel` from the specification in `a`. `channel_in` is then used to initialize `a.dev_mode`, then it is passed to `init`, which sets cfg.dev_mode.

PR #373, opened 2023-04-24
Fix #372: bind to wildcard ip address

### LINKED ISSUE #372: pkasyncio bind to wildcard ip

e-carlin:
In development it is better to bind to the wildcard ip because it is hard to know the network configuration in advance. So, better to listen on all possible interfaces. In environments where this could be problematic the ip will be configured.

### DIFF
--- a/pykern/pkasyncio.py
+++ b/pykern/pkasyncio.py
@@ -90,6 +90,10 @@ def _init():
         return
     _cfg = pkconfig.init(
         debug=(pkconfig.in_dev_mode(), bool, "enable debugging for asyncio"),
-        server_ip=("127.0.0.1", str, "ip to listen on"),
+        server_ip=(
+            "0.0.0.0" if pkconfig.in_dev_mode() else "127.0.0.1",
+            str,
+            "ip to listen on",
+        ),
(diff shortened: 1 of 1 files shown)


pykern/pkasyncio.py
```diff
@@ -90,6 +90,6 @@ def _init():
         return
     _cfg = pkconfig.init(
         debug=(pkconfig.in_dev_mode(), bool, "enable debugging for asyncio"),
-        server_ip=("127.0.0.1", str, "ip to listen on"),
+        server_ip=("0.0.0.0", str, "ip to listen on"),
```
robnagler, line 93: use in_dev_mode to pick the address. In prod, it doesn't make sense to bind to 0.0.0.0.

robnagler (changes requested):

pykern/pkasyncio.py
e-carlin: https://github.com/radiasoft/pykern/compare/a82a0b440ff34556a2228f89392ac84ff5b5dbb0..58dbf6d26f4401b1c88a4f90560e16deb4cd1230

robnagler (approved):

PR #382, opened 2023-07-05
Fix #381: pin urllib to a version that works on centos7

### LINKED ISSUE #381: urlllib3 2.0 dropped support for openssl < 1.1.1
https://github.com/urllib3/urllib3/issues/2168

Our build server runs Centos7 comes with openssl 1.0.2k-fips. I updated python on it to 3.9.15. Running any code that depends on urllib3 raised this error:

```
ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'OpenSSL 1.0.2k-fips  26 Jan 2017'. See: https://github.com/urllib3/urllib3/issues/2168
```

So, pin to a version < 2 until we update OS.

e-carlin:
urllib3 version 2.0 and above only supports openssl > 1.1.1. Centos7 comes with openssl version 1.0.2 so we need to pin urllib3 to a version that works with that openssl version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,12 +6,9 @@
 """
 from pykern.pksetup import setup
 
-setup(
-    name="pykern",
-    description="Python application support",
-    author="RadiaSoft LLC",
-    author_email="pip@pykern.org",
-    install_requires=[
+
+def _requires():
+    r = [
         "argh>=0.26",
         "black~=22.12",
         "future>=0.14",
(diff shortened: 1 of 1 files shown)


robnagler:
We should test for this particular case and use a different version on C7.

robnagler (changes requested):
Test for c7 (or urllib version, but that's tricky, because needs to run on macos too).

e-carlin:
Fixed https://github.com/radiasoft/pykern/pull/382/commits/c07dfebe7bf570d43f26fb1b7ab1b95d5a2e405f .

robnagler (approved):
See my changes

setup.py
```diff
@@ -40,7 +37,42 @@
...
+
+    Adapted from urllib3's checks for ssl
+    https://github.com/urllib3/urllib3/blob/2ac40569acb464074bdc3f308124d781d6aa0860/src/urllib3/__init__.py#L28
+    """
+
+    def _ssl_ok(ssl):
+        return (
+            # python ssl may not be compiled with OpenSSL (ex LibreSSL on MacOS).
+            # https://github.com/urllib3/urllib3/issues/3020
+            getattr(ssl, "OPENSSL_VERSION", "").startswith("OpenSSL ")
```
e-carlin, line 54: Why make this a getattr? Wouldn't it be an ImportError for it to be undefined https://github.com/python/cpython/blob/v3.9.17/Lib/ssl.py#L101 ?

robnagler: `ssl` is defined. `OPENSSL_VERSION` may not be, depending on which version of `ssl` is being installed. If it  doesn't have an `OPENSSL_VERSION`, the new version of `urllib3` will fail to run (with an undefined attr) so we have to fall back.

PR #384, opened 2023-07-22
Fix #383 export pkcollections.object_pairs_hook

### LINKED ISSUE #383: export object_pairs_hook

robnagler:


Benbenbenin0 (approved):
Approving with no comment.

Observed:
- Specify 'else' as 'PKDictNameError' in function comment, more clear
- Moved `object_pairs_hook()` out of parent function.

What is the "exporting" here? Moving the function out from under the parent so it can be imported elsewhere?

robnagler:
> What is the "exporting" here? 

Making "public" is probably a better term. The function is needed by https://github.com/radiasoft/sirepo/pull/6092. 

In Python, "export" really only has a meeting for `__all__`, which is not something we use, because we don't `import *`.

At RS, any attribute that doesn't begin with an underscore is public and implicitly exported/importable.

Issue #62, opened 2020-03-06
Remove tox


robnagler:
Tox is too bulky for what we need, and we'd like to be able to test the testing files that projex creates. Tox is weakly bound (no global files) so just need to modify pksetup.py

robnagler:
pyproject.toml doesn't include tox

Issue #37, opened 2019-04-06
python setup.py <anything> is deprecated


robnagler:
https://github.com/pypa/setuptools/issues/1684 talks about deprecating `python setup.py test` and all other arguments. The plan is to use tox, which we use for testing, but it's unwieldy for other things, like deployment.

Migrate the code out of pysetup into another tool that supports deployment, documentation, etc.

Read the docs build process should be looked, because it calls python setup.py. Perhaps there's a way to just call an arbitrary command or have setup.py do that instead. see https://github.com/rtfd/readthedocs.org/tree/master/readthedocs/doc_builder

See https://github.com/jaraco/skeleton

jeinstei:
Given this deprecation, doing a local user install of a module wouldn't work unless the pip metadata was somehow included as well?

In short, pykern-enabled modules REQUIRE installation using pip? I haven't seen anything in a lot of the package guides on how to programaticaly block this using functions in setup.py, but there has to be some way.

robnagler:
The point of this issue is to move out testing and doc generation from setup.py. Not a rush, because I doubt it will come through.

Installation via pip is not required. Just make sure that `python setup.py` is run at some point.

robnagler:
https://github.com/radiasoft/pykern/issues/466

PR #388, opened 2023-08-14
Fix #387 pkdebug fmt spec

### LINKED ISSUE #387: pkdebug allow format string specifiers
This should work `pkdlog("{:.2f}", .1)`

robnagler:
I removed the running of python setup.py in projex_test.py. It is not something we use, and it was failing weirdly. We need something else which #62 and #37 address.

### DIFF
--- a/pykern/pkdebug.py
+++ b/pykern/pkdebug.py
@@ -80,12 +80,12 @@ value of output is ``$PYKERN_PKDEBUG_OUTPUT``.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 
 """
-from __future__ import absolute_import, division, print_function
 from pykern import pkcompat
 from pykern import pkconfig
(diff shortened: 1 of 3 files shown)


pykern/pkdebug.py
```diff
@@ -665,6 +665,11 @@ def _string(value):
                     if depth > 0:
                         return "'" + s + "'"
                     return s
+                if depth <= 0 and isinstance(
+                    obj, (int, complex, float, decimal.Decimal)
```
e-carlin, line 669: isinstance(obj, numbers.Number) ?

e-carlin (changes requested):

pykern/pkdebug.py
robnagler: bbe5b4b

e-carlin (approved):

PR #392, opened 2023-08-29
Fix #386 readthedocs

### LINKED ISSUE #386: 20240131 Update readthedocs generation
Precipitated by [this notification](https://blog.readthedocs.com/drop-support-system-packages/). Investigate how they currently generate this documentation vs the new method and make changes accordingly.

See also radiasoft/pykern#37

- [x] radiasoft/pykern#429
- [x] radiasoft/pykern#393
- [x] https://github.com/radiasoft/download/issues/526
- [ ] radiasoft/devsecops#521

To clarify a bit, here's what readthedocs is expecting in the repo of interest:
- .readthedocs.yml at the root level, minimally:
```yml
version: 2

build:
  os: ubuntu-22.04
  tools:
    python: "3.9"
  jobs:
    pre_build:
     - <commands to run>

sphinx:
  configuration: docs/conf.py

python:
    install:
    - requirements: docs/requirements.txt
```
- docs/conf.py
- docs/requirements.txt

In the current state of PR radiasoft/pykern#392, "commands to run" is `python setup.py install --force` (which readthedocs used to do automatically). That's done to run pksetup, which generates the conf.py above; accordingly, requirements.txt contains pykern. The purpose of radiasoft/pykern#393 is to take pksetup out of the process and the purpose of https://github.com/radiasoft/download/issues/526 is to make "commands to run" something like `curl https://radia.run | bash -s readthedocs`

mkeilman:
This addresses the recent changes to how [readthedocs](https://docs.readthedocs.io/en/stable/index.html) runs. The big changes are:
- Everything is now governed by `.readthedocs.yml`
- Extra installs must live in `docs/requirements.txt`. For pykern and its cousins the minimal requirement is pykern itself so that pksetp runs
- `setup.py` used to be run automatically ad part of the readthedocs build. Now we must tell it to do so (see the yml file)

We may still need to set up pr triggers because a RTD build has not been run for pykern in almost 5 years.

Once this is confirmed and merged we can roll it out to other repos and modify projex.


robnagler:
@mkeilman is this waiting on me?

mkeilman:
> @mkeilman is this waiting on me?

Do we want to wait for #393 etc.? Some of this code will change to accommodate it.

robnagler:
Yes, I also think we want to automate the process in the readthedocs.yml with a downloader instead of a direct call to pksetup.py (or any python utility). This will give us a chance to fix things more easily in requirements.yml without having to redistribute requirements.yml. I think also the requirements.yml could be simpler, too.

robnagler:
@mkeilman progress?

PR #395, opened 2023-09-11
Fix #394 added pkcompat.zip_strict

### LINKED ISSUE #394: pkcompat.zip_strict

robnagler:


pykern/pkcompat.py
```diff
@@ -99,3 +114,35 @@ def _assert_type(value, typ):
...
+
+
+def _zip_strict(iterables):
+    """Code from https://stackoverflow.com/a/69485272"""
+
+    def _first_tail():
+        nonlocal first_stopped
+        first_stopped = True
+        return
+        yield
```
e-carlin, line 133: How is this reached if we return before?

pykern/pkcompat.py
```diff
@@ -99,3 +114,35 @@ def _assert_type(value, typ):
...
+
+    def _zip_tail():
+        if not first_stopped:
+            raise ValueError("first iterable is longer")
+        for _ in itertools.chain.from_iterable(rest):
+            raise ValueError("first iterable is shorter")
+            yield
```
e-carlin, line 140: How is this reached if we raise before?

pykern/pkcompat.py
```diff
@@ -99,3 +114,35 @@ def _assert_type(value, typ):
                 type(value),
             ),
         )
+
+
+def _version_ok(major, minor):
```
e-carlin, line 119: Is there a reason to prefer this over the code below? I've seen more of the code below (not that that means much...)
```python
def _version_ok(major, minor)
    return sys.version_info >= (major, minor)
```

robnagler: Thanks. I didn't know what I was doing.

pykern/pkcompat.py
robnagler: The yield just forces it to be a coroutine. It's some pretty tricky code.

pykern/pkcompat.py
robnagler: Yield forces it to be a coroutine (generator).

pykern/pkcompat.py
robnagler: 806acaa

robnagler:
I spent a little time trying to figure out how this magic little function works:

```py
def _zip_strict(iterables):
    """Code from https://stackoverflow.com/a/69485272"""

    def _first_tail():
        nonlocal first_stopped
        first_stopped = True
        return
        yield

    def _zip_tail():
        if not first_stopped:
            raise ValueError("first iterable is longer")
        for _ in itertools.chain.from_iterable(rest):
            raise ValueError("first iterable is shorter")
            yield

    if len(iterables) < 2:
        return zip(*iterables)
    first_stopped = False
    iterables = iter(iterables)
    first = itertools.chain(next(iterables), _first_tail())
    rest = list(map(iter, iterables))
    return itertools.chain(zip(first, *rest), _zip_tail())
```

Some definitions:
- `next(any_iterator)` must return `StopIteration` after it was already raised
- `iter` on an iterator returns self
- `zip` doesn't require an iterator 
- `zip` stops iterating the rest of its arguments once an argument is exhausted. This is similar to `in`, btw.
- `chain` concatenates a sequence of iterables
- `from_iterable` concatenates  a single argument of iterables 

Algorithm:
- `first` has a "watcher" (`_first_tail`) that is an empty generator (forced by the `yield`) 
- `rest` has no watchers
- `zip` has a watcher (`_zip_tail`) that checks:
  -  if `_first_tail` did not execute, then `rest` was shorter than `iterable[0]`
  - Otherwise, check if there's nothing left in `rest`, that is, `next` on each element of `rest` must return `StopIteration`

Notes:
- It's important that `rest` be iterators, not iterables. That was confusing to me. Only iterators report `StopIteration`, not sequences. 
- `zip` does not care if something is a sequence or an iterator. It will call `iter` on all its arguments. 
- I didn't understand why the `map` had to call `iter`: it's to ensure that `from_iterable`, which also calls `iter`, doesn't start from the beginning. 
- The algorithm above is highly efficient, because `iter` returns self on an iterator. It's a no-op, and itertools is written in C.
- All the work is done by zip until the last iteration when `_first_tail` and `_zip_tail` execute once. No matter how big the iterables are, the code will run almost linearly (as shown [in the graph](Code from https://stackoverflow.com/a/69485272)).

HTH.

pykern/pkcompat.py
e-carlin: Oh very interesting. Thanks.

pykern/pkcompat.py
```diff
@@ -8,11 +8,12 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
```
rorour, line 8: Line 2: `Python version compatibility routines` seems like a more accurate description now

rorour (approved):
Interesting

pykern/pkcompat.py
robnagler: a2f9f4e2ba9faf60695c181aba2baa216a6cc1f3

PR #408, opened 2023-10-13
Fix #406 added parse_int to pkjson.load_any

### LINKED ISSUE #406: pkjson support standard better on numbers
Sirepo uses JSON to store numbers. There is a problem value `"verticalFocalLength": 100000000000000000000` in an SRW lens, which is 1e20, which is larger than 1e16. This note explains the problem. The solution needs to be overriding how the json decoder works.

Python supports `int` objects with arbitrary magnitude. This is incompatible with JSON, which has a vague definition of [Numbers](https://datatracker.ietf.org/doc/html/rfc8259#section-6): 

> Since software that implements IEEE 754 binary64 (double precision) numbers [IEEE754](https://datatracker.ietf.org/doc/html/rfc8259#ref-IEEE754) is generally available and widely used, good interoperability can be achieved by implementations that expect no more precision or range than these  provide, in the sense that implementations will approximate JSON numbers within the expected precision.

I think we should assume JSON, like Javascript, does not distinguish between integers and floats. All [Numbers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#number_type) are 64 bit IEEE 754 values. Number defines [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) and [MIN_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER), which are 53 bits or just shy of 16 decimal digits. 

Javascript encodes 1e20 (problematic number) in JSON as:

```sh
> JSON.stringify(1e20)
'100000000000000000000'
```

Python's JSON interpreter uses Python parsing rules instead of JSON standard parsing rules. It interprets a number without a `.` as an `int`, which JSON has no concept of. Instead, once parse_int converts the number, it should a float if the number is larger than MAX_SAFE_INTEGER or smaller than MIN_SAFE_INTEGER.

robnagler:
- Sets stronger limit (64 chars) than Python's parse_int (4300 chars)

### DIFF
--- a/pykern/pkcollections.py
+++ b/pykern/pkcollections.py
@@ -329,6 +329,10 @@ Dict = PKDict
 DictNameError = PKDictNameError
 
 
+_JSON_INT_MAX = 2**53 - 1
+_JSON_INT_MIN = -(2**53) + 1
+
+
 def json_load_any(obj, *args, **kwargs):
     """Parse json with `PKDict` for object pairs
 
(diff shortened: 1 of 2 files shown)


rorour (approved):
Looks good to me

Issue #74, opened 2020-06-10
pkconfig should catch errors, wrap them, and rethrow


robnagler:
when a config variable is in error, pkconfig should document where it got the value and what the value's name is.

robnagler:
I think `_iter_decls` is the only place that needs to be fixed. Right around resolution.

PR #411, opened 2023-10-18
Fix #405: init db dir

### LINKED ISSUE #405: Generalize srdb._init_root
[_init_root](https://github.com/radiasoft/sirepo/blob/51f5712a9f812ce9abab9a0ffccef80a2613ad2c/sirepo/srdb.py#L51) does a lot of work, which could be generalized to something like pykern.util.init_db_dir and .cfg_db_dir (cfg parser).

gurhar1133:
See [related sirepo PR](https://github.com/radiasoft/sirepo/pull/6453)

### DIFF
--- a/pykern/util.py
+++ b/pykern/util.py
@@ -25,7 +25,7 @@ _cfg = None
 
 
 #: where job db is stored under srdb.root
-_SUPERVISOR_DB_SUBDIR = "supervisor-job"
+# _SUPERVISOR_DB_SUBDIR = "supervisor-job"
 
 
 # def proprietary_code_dir(sim_type):
@@ -41,10 +41,10 @@ def root():
     return _root or _init_root()
 
 
-def supervisor_dir():
-    """Directory for supervisor job db"""
+# def supervisor_dir():
+#     """Directory for supervisor job db"""
 
-    return root().join(_SUPERVISOR_DB_SUBDIR)
+#     return root().join(_SUPERVISOR_DB_SUBDIR)
 
 
 def _init_root():
@@ -61,10 +61,9 @@ def _init_root():
     _cfg = pkconfig.init(
         root=(None, _cfg_root, "where database resides"),
     )
-    pkdp("\n\n\n _cfg={}", _cfg)
-    # _root = _cfg.root
-    # if _root:
-    #     return _root
+    _root = _cfg.root
+    if _root:
+        return _root
     assert pkconfig.in_dev_mode(), "SIREPO_SRDB_ROOT must be configured except in dev"
     r = (
         pkio.py_path(
@@ -73,17 +72,11 @@ def _init_root():
         .dirpath()
         .dirpath()
     )
-
-    # TODO (gurhar1133): a check for if we are in dev_mode?
-
-    # assert 0, f"r={r}"
     # Check to see if we are in our dev directory. This is a hack,
     # but should be reliable.
     if not r.join("setup.py").check():
         # Don't run from an install directory
         r = pkio.py_path(".")
+    pkdp("r={}, _cfg={}", r, _cfg)
     _root = pkio.mkdir_parent(r.join(_DEFAULT_ROOT))
     return _root
-
-if __name__ == "__main__":
-    _init_root()
--- a/pykern/util.py
+++ b/pykern/util.py
@@ -21,20 +21,20 @@ _root = None
 _cfg = None
 
 #: subdir of where proprietary codes live
-_PROPRIETARY_CODE_DIR = "proprietary_code"
+# _PROPRIETARY_CODE_DIR = "proprietary_code"
 
 
 #: where job db is stored under srdb.root
 _SUPERVISOR_DB_SUBDIR = "supervisor-job"
 
 
-def proprietary_code_dir(sim_type):
-    """Directory for proprietary code binaries
+# def proprietary_code_dir(sim_type):
+#     """Directory for proprietary code binaries
 
-    Args:
-        sim_type (str): not validated code name
-    """
-    return root().join(_PROPRIETARY_CODE_DIR, sim_type)
+#     Args:
+#         sim_type (str): not validated code name
+#     """
+#     return root().join(_PROPRIETARY_CODE_DIR, sim_type)
 
 
 def root():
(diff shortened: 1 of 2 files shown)


gurhar1133:
Thoughts on a test (if necessary)? init a run dir, check that the path is what it should be (accounting for the difference between dev and ci) and remove the run dir?

pykern/util.py
```diff
@@ -0,0 +1,62 @@
...
+def cfg_db_dir(db_root_env_name):
+    """Generate pkconfig parser function with errors for db's env var name
+
+    Args:
+        db_root_env_name (str): env var name for run dir root
+
+    Returns:
+        function: parser function for pkconfig.init
+    """
+    _NOT_ABS_ERROR = "{}: " + f"{db_root_env_name} must be absolute"
```
robnagler, line 27: Good solution, but db_root_env_name shouldn't need to be passed. Rather https://github.com/radiasoft/pykern/issues/74 should be fixed to always print the key on errors. This would simplify this code, since the error messages wouldn't need to print the key.

pykern/util.py
```diff
@@ -0,0 +1,62 @@
...
+    _NO_DIR_ERROR = "{}: " + f"{db_root_env_name} must be a directory and exist"
+
+    def cfg_db_dir_parser(v):
+        """Config value or root package's parent or cwd with `_DEFAULT_ROOT`"""
```
robnagler, line 31: The value shouldn't need to be printed either. Rather, this should be handled by the https://github.com/radiasoft/pykern/issues/74

pykern/util.py
```diff
@@ -0,0 +1,62 @@
...
+def init_db_dir(package_object):
+    """Initialize run dir
+
+    Args:
+        package_object (object): python object who's root_package will hold the run dir
+
+    Returns:
+        py.path.local: absolute path to run dir
+    """
+    r = (
```
robnagler, line 50: Need to assert dev mode. This should be called something like `dev_run_dir`. It's more than "db" in that it is the root of where all run artifacts are held.

pykern/util.py
```diff
@@ -0,0 +1,62 @@
...
+        pkio.py_path(
+            sys.modules[pkinspect.root_package(package_object)].__file__,
+        )
+        .dirpath()
+        .dirpath()
+    )
+    # Check to see if we are in our dev directory. This is a hack,
+    # but should be reliable.
+    if not r.join("setup.py").check():
```
robnagler, line 59: Check for pyproject.toml, too. We'll be moving to that soonish.

pykern/util.py
```diff
@@ -0,0 +1,62 @@
...
+        # Don't run from an install directory
+        r = pkio.py_path(".")
```
robnagler, line 61: Technically py_path(".") is equivalent to py_path().

robnagler (changes requested):

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+from pykern import pkio
+from pykern.pkdebug import pkdc, pkdexc, pkdlog, pkdp
+import os.path
+import sys
+
+
+_DEFAULT_ROOT = "run"
+
+
+def cfg_db_dir_parser(value):
```
robnagler, line 18: Name this `cfg_absolute_dir`. It is any path that is absolute and a directory.

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+    """Config value or root package's parent or cwd with `_DEFAULT_ROOT`"""
```
robnagler, line 19: The comment is incorrect. Update comment including Args and Returns.

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+    if not os.path.isabs(value):
+        pkconfig.raise_error("must be absolute")
+    if not os.path.isdir(value):
+        pkconfig.raise_error("must be a directory and exist")
+    return pkio.py_path(value)
+
+
+def dev_run_dir(package_object):
+    """Initialize run dir
```
robnagler, line 28: Comment like this is not useful. "Initialize" is incorrect. Describe what it does.

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Args:
+        package_object (object): python object who's root_package will hold the run dir
```
robnagler, line 31: who's whose

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    # Check to see if we are in our dev directory. This is a hack,
+    # but should be reliable.
+    if not _check_for_files(r, ("setup.py", "pyproject.toml")):
+        # Don't run from an install directory
+        r = pkio.py_path()
+    return pkio.mkdir_parent(r.join(_DEFAULT_ROOT))
+
+
+def _check_for_files(root, files):
```
robnagler, line 52: make this a nested function

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+    for f in files:
+        if not root.join(f).check():
```
robnagler, line 54: I don't think this check is correct. You might consider using a filter function for this, e.g. `any`.

pykern/util.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    assert pkconfig.in_dev_mode(), "run dir root must be configured except in dev"
+    r = (
+        pkio.py_path(
+            sys.modules[pkinspect.root_package(package_object)].__file__,
+        )
+        .dirpath()
+        .dirpath()
+    )
+    # Check to see if we are in our dev directory. This is a hack,
```
robnagler, line 44: This comment is not very good. I know I wrote it. Look for files at the root of a python project that are not installed when the package is installed. There are assumptions (above) about where this path is so we assume a particular file structure. This should be noted in the function's description.

robnagler (changes requested):

gurhar1133:
Another look @robnagler ?

e-carlin (changes requested):

pykern/util.py
```diff
@@ -0,0 +1,66 @@
...
+
+    assert pkconfig.in_dev_mode(), "run dir root must be configured except in dev"
+    r = (
+        pkio.py_path(
+            sys.modules[pkinspect.root_package(package_object)].__file__,
+        )
+        .dirpath()
+        .dirpath()
+    )
+    a = pkio.py_path(
```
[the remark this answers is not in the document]
gurhar1133: Oops. That was for debugging but left it in on accident.

e-carlin (approved):

PR #412, opened 2023-10-19
Fix #74: wrap resolution error

### LINKED ISSUE #74: pkconfig should catch errors, wrap them, and rethrow

gurhar1133:
Fix for [#411 ](https://github.com/radiasoft/pykern/pull/411)


pykern/pkconfig.py
```diff
@@ -730,7 +730,10 @@ def _iter_decls(decls, res):
         if d.group:
             r[kp] = PKDict()
             continue
-        r[kp] = _resolver(d)(k, d)
+        try:
+            r[kp] = _resolver(d)(k, d)
+        except Exception as e:
+            raise_error(f"Error: {e}, key={kp}, value={decls[k]}")
```
robnagler, line 736: `exception={e} key=...` 

We want to reraise here but augment the existing error. See [fconf.exception_reason](https://github.com/radiasoft/pykern/blob/9fc4608c1b3094b08b227245477081758945778d/pykern/fconf.py#L205) for an example.

Also, add a test so that we can test-document this.

robnagler (changes requested):

pykern/pkconfig.py
e-carlin: Could we use https://docs.python.org/3/library/exceptions.html#exception-context ?

robnagler (approved):

PR #416, opened 2023-11-08
Fix #415 change regex to include any user + added to test case

### LINKED ISSUE #415: github_bkp silence errors related to missing wiki for non radiasoft repos
In https://github.com/radiasoft/pykern/pull/376 we fixed silencing of errors for repos that have no wiki. But, the [regex test](https://github.com/radiasoft/pykern/blob/aff64cd690713836e84acd4b75e18590afa15f5b/pykern/pkcli/github.py#L31) we use is too restrictive. It checks for only "radiasoft" wikis. We need to handle non-rs wikis too (ex biviosoftware). 

Let's relax the regex to allow for any org/user. I think we can change the regex to look for `[\w-\.]+` for both the username and repo name https://stackoverflow.com/a/59082561.

@Benbenbenin0 not the highest priority issue but can you get to it in the next week or so?

Benbenbenin0:
@e-carlin my guess is we don't want to use the biviosoftware repos I've added since they aren't test repos (wiki could change). Any suggestions for non-radiasoft test repos?

Also I added underscore to suggested regex from issue since also allowable for github usernames.

### DIFF
--- a/pykern/pkcli/github.py
+++ b/pykern/pkcli/github.py
@@ -28,7 +28,7 @@ import time
 _GITHUB_HOST = "github.com"
 _GITHUB_URI = "https://" + _GITHUB_HOST
 _GITHUB_API = "https://api." + _GITHUB_HOST
-_WIKI_ERROR_OK = r"fatal: repository 'https://github.com/[\w\-\.\_]+/[\w\-\.\_]+.wiki.git/' not found"
+_WIKI_ERROR_OK = r"fatal: repository 'https://github.com/[-/\w\.]+.wiki.git/' not found"
 _RE_TYPE = type(re.compile(""))
 _MAX_TRIES = 3
 _TEST_REPOS = [
--- a/pykern/pkcli/github.py
+++ b/pykern/pkcli/github.py
@@ -28,12 +28,15 @@ import time
 _GITHUB_HOST = "github.com"
 _GITHUB_URI = "https://" + _GITHUB_HOST
 _GITHUB_API = "https://api." + _GITHUB_HOST
-_WIKI_ERROR_OK = (
-    r"fatal: repository 'https://github.com/radiasoft/.*.wiki.git/' not found"
-)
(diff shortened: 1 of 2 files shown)


pykern/pkcli/github.py
```diff
@@ -28,12 +28,15 @@
 _GITHUB_HOST = "github.com"
 _GITHUB_URI = "https://" + _GITHUB_HOST
 _GITHUB_API = "https://api." + _GITHUB_HOST
-_WIKI_ERROR_OK = (
-    r"fatal: repository 'https://github.com/radiasoft/.*.wiki.git/' not found"
-)
+_WIKI_ERROR_OK = r"fatal: repository 'https://github.com/[\w\-\.\_]+/[\w\-\.\_]+.wiki.git/' not found"
```
e-carlin, line 31: [\w matches underscore](https://docs.python.org/3/library/re.html). I think add '/' to the regex and eliminate the need for both of them. Something like:

```python
_WIKI_ERROR_OK = r"fatal: repository 'https://github.com/[-/\w\.]+.wiki.git/' not found"
```

pykern/pkcli/github.py
```diff
@@ -28,12 +28,15 @@
...
 _RE_TYPE = type(re.compile(""))
 _MAX_TRIES = 3
-_TEST_REPOS = ["test-pykern-github", "test-pykern-github-no-wiki"]
+_TEST_REPOS = [
+    ("radiasoft", "test-pykern-github"),
+    ("radiasoft", "test-pykern-github-no-wiki"),
+    ("biviosoftware", "home-env"),
```
e-carlin, line 37: @moellep can you create an repo in biviosoftware? Please name it something like `test-pykern-github-no-wiki` and put just a README.md in the repo

e-carlin (changes requested):

pykern/pkcli/github.py
moellep: I've created https://github.com/biviosoftware/test-pykern-github-no-wiki now

Benbenbenin0: Added to test

e-carlin (approved):
perfect

PR #422, opened 2023-12-06
Fix #418: remove argh pin

### LINKED ISSUE #418: remove argh pin
We pinned argh because an update caused breakage https://github.com/radiasoft/pykern/pull/414 . Remove this pin and update our code accordingly.

e-carlin:
In argh v0.30.0 there was a breaking change in the way argh mapped function arguments to CLI arguments. For example, if you had a function like `def f(x, bar=None)` prior to v0.30.0 bar was an optional keyword argument which means on the CLI it was set with a flag (--bar BAR). In v0.30.0 and up bar became a positional arg with a default. So, you could no longer invoke the function from the CLI with --bar set.

To fix this we now set the NameMappingPolicy to BY_NAME_IF_HAS_DEFAULT
 which retains the closest approximation with the old behavior.


pykern/pkcli/__init__.py
```diff
@@ -25,6 +25,7 @@
 """
 from __future__ import absolute_import, division, print_function
```
robnagler, line 26: remove

pykern/pkcli/__init__.py
```diff
@@ -162,6 +171,24 @@ def main(root_pkg, argv=None):
...
 
+def _argh_name_mapping_policy():
+    """Set the argh name_mapping_policy.
+
+    In v0.30.0 argh changed the policy for mapping function args to
+    CLI args. For versions 0.30.0 and above this sets the policy to
+    BY_NAME_IF_HAS_DEFAULT which most closely matches the behavior
+    prior to v0.30.0.
+    """
+    from pykern.pkcollections import PKDict
```
robnagler, line 182: Do not import PKDict here. We try to avoid imports. I realize it can be imported, but it's better not, since this is before the pkcli has be initialized.

tests/pkcli/argh_test.py
```diff
@@ -0,0 +1,22 @@
...
+    from pykern.pkunit import pkre
+    from pykern.pkcollections import PKDict
+    import subprocess
+
+    p = "p"
+    k = "k"
+    pkre(
+        str(PKDict(positional_arg=p, keyword_arg=k)),
+        subprocess.check_output(
+            ("pykern", "argh_test", "kwarg-to-positional", p, "--keyword-arg", k),
```
robnagler, line 19: This will not test the right pykern. It will test pykern that's installed. This needs to test the pykern in the subdirectories. It's tricky to do this. Rather, you can call pkcli directly: `pkcli.main("pykern")`. You setup argv. Look at other pkcli tests.

pykern/pkcli/argh_test.py
```diff
@@ -0,0 +1,25 @@
+"""Test pkcli to handle argh breaking changes
```
robnagler, line 1: This should be unit test data. See below.

pykern/pkcli/__init__.py
```diff
@@ -162,6 +171,24 @@ def main(root_pkg, argv=None):
     return 0
 
 
+def _argh_name_mapping_policy():
+    """Set the argh name_mapping_policy.
```
robnagler, line 175: `Return name_mapping_policy based on feature test of NameMappingPolicy`

tests/pkcli/argh_test.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+"""pytest for `pykern.pkcli.argh_test`
```
robnagler, line 2: We don't need to mention pytest. It is a test for pkcli arg parsing.

robnagler (changes requested):

tests/pkcli/argh_test.py
robnagler: python-ci-run installs, but that's not classical CI behavior. Typically, you test before installing, e.g. make test, make install, not make install, make test. 

[Modify sys.path in the test](https://github.com/radiasoft/pykern/blob/d279e4ddccdf48165ddc1886f15ef2521b211efc/tests/pkcli_test.py#L122) so that it can find the pkcli modules.

e-carlin: I see `pykern ci run` enforces the cwd is the top-level of the package under test so it is added to PYTHONPATH implicitly. 

Still if we were to run `pykern test path/to/test` which doesn't have this enforcement then I think we still test the installed version and not whatever version we are running the test file from.

I tried this by installing pykern from master. Then cloning pykern again, checking out this branch, chaning the test to have `import pykern.pkcli.argh_test`, and invoking `pykern test /path/to/pykern`. It failed with a ModuleNotFoundError

e-carlin: I gave it a shot 376dc37 . Seems goofier than calling subprocess and from what I can tell unnecessary since it looks like always (or close enough) we test the installed package.

pykern/pkcli/__init__.py
e-carlin: 7ff4012

pykern/pkcli/__init__.py
e-carlin: 7ff4012

tests/pkcli/argh_test.py
e-carlin: 7ff4012

tests/pkcli/argh_test.py
robnagler: `pykern` has to be installed to run it as a command. That's distinct from what is being tested. For development, we use egg links (or easy-install.pth, probably). That's fine.

`pkcli_test.py` does this work already so not sure why you are duplicating it in another test. Just add argh_test.py to pkcli_data and add a case function.

For "modern" CI and development, it makes sense to have it be installed so we don't have to muck with paths everywhere, e.g. to run pykern. However, there's a principle here regarding testing. Specifically, argh_test.py is data, and we should avoid adding it as an installed pkcli. This mucks up autocomplete. nersc_test is something different, and has to be installed (because it is for NERSC's CI of our docker images). Sometimes it is difficult to separate unit test data, but most times it isn't.

You can verify pkcli_test.py tests the local file with:

```sh
# this should not be found
grep pykern ~/.pyenv/versions/py3/lib/python3.9/site-packages/easy-install.pth
rm -rf ~/.pyenv/versions/py3/lib/python3.9/site-packages/pykern*
# should get not found
pykern
cd ~/src/radiasoft/pykern/tests
# should pass
PYTHONPATH=~/src/radiasoft/pykern/tests pykern test pkcli_test.py
```

robnagler (changes requested):
see https://github.com/radiasoft/pykern/pull/422#discussion_r1417742112

tests/pkcli/argh_test.py
e-carlin: The commands you gave don't work for me:
```
$ grep pykern ~/.pyenv/versions/py3/lib/python3.9/site-packages/easy-install.pth
/home/vagrant/src/radiasoft/pykern
$ pip uninstall pykern
<snip>
  Successfully uninstalled pykern-20231024.210332
$ grep pykern ~/.pyenv/versions/py3/lib/python3.9/site-packages/easy-install.pth
$ rm -rf ~/.pyenv/versions/py3/lib/python3.9/site-packages/pykern*
$ pykern
bash: pykern: command not found
$ cd ~/src/radiasoft/pykern/tests
tests$ PYTHONPATH=~/src/radiasoft/pykern/tests pykern test pkcli_test.py
bash: pykern: command not found
```

Regardless I made the changes 1afbd34

pykern/pkcli/argh_test.py
e-carlin: 1afbd34

tests/pkcli/argh_test.py
robnagler: he he. You didn't follow the instructions to the letter. Since it's pyenv, I didn't include `pip uninstall pykern`. This removes `pykern` the command, which is a proxy in pyenv. This should work after a pip uninstall:

`PYTHONPATH=~/src/radiasoft/pykern/tests python ~/src/radiasoft/pykern/pykern/pykern_console.py test pkcli_test.py`

tests/pkcli_data/arghparsing/pkcli/__init__.py
```diff
@@ -0,0 +1 @@
+#
```
robnagler, line 1: These are unnecessary. You can just use `p1`. This simplifies the test data. Sorry to be pedantic, but I think for future test-learners, this is important. The argh_test.py is clear enough.

robnagler (changes requested):

tests/pkcli/argh_test.py
e-carlin: They didn't give enough detail so I went rogue. How do I achieve "# this should not be found" w/o an uninstall? Remove it from the file by hand? Something else?

I tried removing it by hand. Now I get 
```
tests (418-argh-pin)$ PYTHONPATH=~/src/radiasoft/pykern/tests pykern test pkcli_test.py
  File "/home/vagrant/.pyenv/versions/py3/bin/pykern", line 33, in <module>
    sys.exit(load_entry_point('pykern', 'console_scripts', 'pykern')())
  File "/home/vagrant/.pyenv/versions/py3/bin/pykern", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/metadata.py", line 542, in distribution
    return Distribution.from_name(distribution_name)
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/metadata.py", line 196, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: pykern
```

e-carlin: I ran just the command you gave me 
```
PYTHONPATH=~/src/radiasoft/pykern/tests python ~/src/radiasoft/pykern/pykern/pykern_console.py test pkcli_test.py case=argh
  File "/home/vagrant/src/radiasoft/pykern/pykern/pykern_console.py", line 13, in <module>
    from pykern import pkcli
ModuleNotFoundError: No module named 'pykern'
```

robnagler: Yes, that's it. `pykern` (pykern_console.py) the executable is still in your path, but it is just a simple program that loads pykern pkcli, which does all the work.

tests/pkcli_data/arghparsing/pkcli/__init__.py
robnagler: These are good points. I only have opinions that do not mean much, but I offer them anyway. :)

`p1` could be renamed `package1` if that's clearer to you.

One question you raise is whether each case should be in its own package. That I think is perhaps valid or overkill. I don't find it useful to have empty files, e.g. argparsing/__init__.py`. This implies that there's a module named argparsing in use somewhere. That was my confusion when I first read it, and I though you were mocking it in the test.

The case name itself `test_argh_argument_parsing` is what's relevant to me. It's the first place anybody should look to see if there's a test for this case, e.g. search on `argh` in the test files for pkcli. Alternatively, it might be in a string for pkeq or something. We don't usually do that, because the test case names (to me at least) should be clear as to what they are testing. In any case, it will be in the test file. 

The test data itself is very tricky to understand without understanding the test code and the code that is being tested. It wasn't clear to me why, for example, you had a `pkeq(str(PKDict...)` until I saw the result. One could argue here (and I didn't at first :) that the use of `str` is implicitly coupled with the way pkcli main behaves, which is actually not the purpose of the test. If, for example, you simply stringified the result instead of returning a dict, then it would be clearer (at least to me) that the two values are the same. There's an awful lot to unpack here, but it's something that's subtle about testing: be careful about including too much in the test.

> I end up at functions like pkcli_test._dev. Another short name that tells me nothing. 

Unfortunately, that's just it. I wouldn't have known a good name for that function. A longer name might be fine, but there's a lot of context in these functions. The word `dev` is overloaded, I agree. `_deviance` would be clearer, but I don't know that it would explain things, but then this code would be harder to read:

```py
    _dev(rp, [], None, all_modules, capsys)
    _dev(rp, ["--help"], None, all_modules, capsys)
    _dev(rp, ["conf1"], SystemExit, r"cmd1,cmd2.*too few", capsys)
    _dev(rp, ["conf1", "-h"], SystemExit, r"\{cmd1,cmd2\}.*commands", capsys)
    if six.PY2:
        _dev(rp, ["not_found"], None, r"no module", capsys)
    else:
        _dev(rp, ["not_found"], ModuleNotFoundError, None, capsys)
    _dev(rp, ["conf2", "not-cmd1"], SystemExit, r"\{cmd1\}", capsys)
```

Perhaps a for loop would help, or not, because cases failures would be harder to track.

I'm not dying on any hills here. Just saying my opinions. The fundamental one is that testing is very hard to get right.

tests/pkcli/argh_test.py
robnagler: Yes, I meant to resolve. If you go back to https://github.com/radiasoft/pykern/pull/422#discussion_r1417742112, and do those steps, it will verify that there's no module installed (ModuleNotFoundError) until you add a PYTHONPATH, which will point to the uninstalled directory.

tests/pkcli_data/arghparsing/pkcli/__init__.py
e-carlin: > p1 could be renamed package1

1000x yes. The name package1 gets me a lot closer to realizing that the name is meaningless. I wish I knew why but with the name p1 I really don't know what to think. When names are hard to understand (p1) my first thought is that I don't understand the code. Not that the name itself is meaningless.

After learning that p1 means package1 I'm less inclined to have a separate package for the argh testing. I still like that it is separated out in the negative case. Also, the arg case feels a bit unique to me since it is a test for dealing with the peculiarities of a release of a dependency and I can imagine a future where we get rid of the test. I think separating it out makes it easier to get rid of.  But, I'm not too sold on any of those ideas.

>  The word dev is overloaded, I agree. _deviance would be clearer

Again, an emphatic yes. As you say, dev is overloaded. Deviance is a long way down the list of names I might guess. And I'd rather not guess at all. deviance actually tells me quite a bit about what is going on. It doesn't paint the whole picture but it gets me started. A relevant example: initially when I wrote the test in here I tried to use _dev because I saw other commands using it that looked like they were doing what I wanted to do. It wasn't until I ran into the assert for a return value of 1 that I realized what was going on. I think (hope) that "_deviance" would have clued me in sooner. 

> One could argue here (and I didn't at first :) that the use of str is implicitly coupled with the way pkcli main behaves, which is actually not the purpose of the test

I think switching to a string is a good idea.

Thanks again for the details.

6b14220

robnagler (approved):
looks great to me :)

PR #425, opened 2023-12-19
Fix #424: is binary

### LINKED ISSUE #424: pkio needs is_binary()

gurhar1133:
Using ideas from the [article](https://eli.thegreenplace.net/2011/10/19/perls-guess-if-file-is-text-or-binary-implemented-in-python) you shared with me @robnagler 

@robnagler you also mentioned that if the file is utf-8 we can assume that it is text. Since ascii is a subset of utf-8 i changed the method of counting how many characters are ascii to counting how may are utf-8.


pykern/pkio.py
```diff
@@ -125,6 +125,30 @@ def has_file_extension(filename, to_check):
     return e in to_check
 
 
+def is_binary(filename, threshold=0.30):
```
robnagler, line 128: I don't think threshold is important. The routine should be magic. If we need to parameterize, it's not magic enough. For example, we could read more data.

is_binary should be a [sans I/O](https://news.ycombinator.com/item?id=12242628) function in pykern.util and called by pkio for a convenient wrapper.

I think the test should be called `is_pure_text` or something like that. `is_binary` has many meanings, e.g. has executable bit set.

pykern/pkio.py
```diff
@@ -125,6 +125,30 @@ def has_file_extension(filename, to_check):
...
+    """Uses heuristics to guess whether file is binary data
+
+    Args:
+        filename (str|py.path.local): file to check
+        threshold (float): allowed percentage of assumed non-textual characters
+
+    Returns:
+        bool: True if file is likely binary data
+    """
+    if threshold >= 1:
```
robnagler, line 138: We don't normally test this. However, if you do, it should not be allowed to be negative either.

pykern/pkio.py
```diff
@@ -125,6 +125,30 @@ def has_file_extension(filename, to_check):
...
+        raise AssertionError(
+            f"threshold={threshold} must be < 1 (represents percentage)"
+        )
+    with open(filename, "rb") as f:
+        b = f.read(512)
+        if not b:
```
robnagler, line 144: close the file at this point, that is,

```py
with open() as f
    b = f.read()
if not b:
```

The conditional on `b` should be robust, e.g. `if (l := len(b)) <= 0`. `l` is used below so might as well do the check around that.

pykern/pkio.py
```diff
@@ -125,6 +125,30 @@ def has_file_extension(filename, to_check):
...
+
+    Returns:
+        bool: True if file is likely binary data
+    """
+    if threshold >= 1:
+        raise AssertionError(
+            f"threshold={threshold} must be < 1 (represents percentage)"
+        )
+    with open(filename, "rb") as f:
+        b = f.read(512)
```
robnagler, line 143: If threshold is desired, then this should be parameterized, too.

pykern/pkio.py
```diff
@@ -125,6 +125,30 @@ def has_file_extension(filename, to_check):
...
+        if not b:
+            return False
+        if b"\x00" in b:
+            return True
+        l = len(b)
+        return (l - len(b.decode("utf-8", "ignore"))) / l > threshold
```
robnagler, line 149: If it can't decode utf8, then it should be binary, i.e. do not ignore errors. The likelihood of truncating at a utf8 boundary is low but real. Try to decode as utf8. If succeeds, then it's text. If it fails, look at the last three bytes to see if there's a utf start character and backup from there, possibly iteratively. 

I am not clear the threshold applies any more now that we have utf8 almost exclusively. We don't have to worry about other codecs. It may be that simply decoding for utf8 is sufficient.

We might add an iso 8859 codec, since that would be easy. I don't think we need to handle much else.

tests/pkio_test.py
```diff
@@ -59,6 +59,16 @@ def test_has_file_extension():
...
 
 
+def test_is_binary():
+    from pykern import pkunit
+    from pykern import pkio
+
+    d = pkunit.data_dir()
+    pkunit.pkeq(False, pkio.is_binary(d.join("empty.dat")))
+    pkunit.pkeq(False, pkio.is_binary(d.join("text.dat")))
+    pkunit.pkeq(True, pkio.is_binary(d.join("binary.dat")))
```
robnagler, line 69: The advantage of doing an sans-i/o implementation would be to not have this type of code. You could simply have byte strings. The binary.dat is too complicated. It will always pass muster. You want to find something that's at the edge of the threshold to test.

robnagler (changes requested):

gurhar1133:
@robnagler I decoupled the i/o from the parsing. Right now there is a test for both of those layers (pkio_test and util_test) would you like all the tests to be in pkio or util or is testing both those layers good?

pykern/util.py
```diff
@@ -65,3 +65,40 @@ def _check_dev_files(root):
...
+            continue
+    return False
+
+
+def _decode_success(data, decode_failure_callback):
+    try:
+        data.decode("utf-8", "strict")
+        return True
+    except UnicodeDecodeError:
+        return decode_failure_callback()
```
robnagler, line 104: I changed some names to be a bit more descriptive. "chunk" implies multiple chunks. "test_size" is suboptimal, but good enough. Possibly "probe_len".

Try to write less code, especially avoiding callbacks unless absolutely necessary.

Note the change in the test on test_size. If value is exactly test_size, just test it, you can't probe ahead anyway.

By going backwards, you avoid the read ahead problem. The test_size is a guess so we can go backwards, and it will be good enough.

```py
def is_pure_text(value, test_size=None):
    def _try(chunk):
        try:
            chunk.decode("utf-8", "strict")
            return True
        except UnicodeDecodeError:
            return False

    if not test_size or len(value) <= test_size:
        return _try(value)
    b = value[:chunk_size]
    # 4 is maximum length of a utf8 char so if a char
    # is truncated by test_size, we need to probe back
    # a bit to find the end of the char.
    for _ in range(4):
        if _try(b):
            return True
        if len(b) <= 1:
            return False
        b = b[:-1]
    return False
```

tests/util_test.py
```diff
@@ -0,0 +1,28 @@
...
+    pkunit.pkeq(True, util.is_pure_text(b""))
+    pkunit.pkeq(True, util.is_pure_text(b"This is example text"))
+    pkunit.pkeq(
+        False,
+        util.is_pure_text(
+            b"\xeb\xbf\xe4\xa0\xe9\xcf\xa5\n\xe7\xbf\xde\x83\xd4\x16\xc0\xd6\xec\xbf\x92\xe6\x84T\xc9 \xe9\xbf"
+        ),
+    )
+    pkunit.pkeq(
+        True, util.is_pure_text(("a" * 511).encode("utf-8") + "¡".encode("utf-8"), 512)
```
robnagler, line 23: no need to test 512 bytes. just set text_size to something shorter that tests the edge cases.

tests/util_test.py
```diff
@@ -0,0 +1,28 @@
...
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+import pytest
+
+
+def test_is_pure_text():
+    from pykern import pkunit
+    from pykern import util
+
+    pkunit.pkeq(True, util.is_pure_text(b""))
```
robnagler, line 14: test `b"\0"`, which will decode properly, but is incorrect. Here's how perl does it:

> The -T and -B tests work as follows. The first block or so of the file is examined to see if it is valid UTF-8 that includes non-ASCII characters. If so, it's a -T file. Otherwise, that same portion of the file is examined for odd characters such as strange control codes or characters with the high bit set. If more than a third of the characters are strange, it's a -B file; otherwise it's a -T file. Also, any file containing a zero byte in the examined portion is considered a binary file. (If executed within the scope of a [use locale](https://perldoc.perl.org/perllocale) which includes LC_CTYPE, odd characters are anything that isn't a printable nor space in the current locale.) If -T or -B is used on a filehandle, the current IO buffer is examined rather than the first block. Both -T and -B return true on an empty file, or a file at EOF when testing a filehandle. Because you have to read a file to do the -T test, on most occasions you want to use a -f against the file first, as in next unless -f $file && -T $file.

tests/util_test.py
```diff
@@ -0,0 +1,28 @@
...
+    pkunit.pkeq(True, util.is_pure_text(b"This is example text"))
```
robnagler, line 15: Don't test for "True" or "False". Rather, use pkok. You may consider wrapping this in a function or a loop to make the test code more readable.

pykern/pkio.py
```diff
@@ -125,6 +127,20 @@ def has_file_extension(filename, to_check):
     return e in to_check
 
 
+def is_pure_text(filename, chunk_size=512):
+    """Uses pykern.util.is_pure_text to guess whether file is pure text
+
+    Args:
+        filename (str|py.path.local): file to check
```
robnagler, line 134: describe chunk_size, but see discussion in util.

pykern/pkio.py
```diff
@@ -125,6 +127,20 @@ def has_file_extension(filename, to_check):
...
+
+    Returns:
+        bool: True if file is likely pure text, false if likely binary
+    """
+    with open(filename, "rb") as f:
+        b = f.read(chunk_size + _POSSIBLE_VALID_CHAR_BYTES)
```
robnagler, line 140: no need for temporary variable. No need for _POSSIBLE_VALID_CHAR_BYTES. Boundary is arbitrary. See discussion below.

pykern/pkio.py
```diff
@@ -125,6 +127,20 @@ def has_file_extension(filename, to_check):
     return e in to_check
 
 
+def is_pure_text(filename, chunk_size=512):
```
robnagler, line 130: util should have the same default value for chunk_size.

robnagler (changes requested):

robnagler:
> Right now there is a test for both of those layers (pkio_test and util_test) would you like all the tests to be in pkio or util or is testing both those layers good?

The test for pkio can be very simple, since the code is simple. Just two cases should be enough.

tests/util_test.py
gurhar1133: I added the "\x00" heuristic check. Would there be value in adding back in the " If more than a third of the characters are strange" heuristic as well?

robnagler: You have to test for more characters than "\x00". Read the perl description again, and figure out what those might be.

robnagler: Also the check has to happen after the utf8 check. utf8 allows for "\x0" so check the result in try after the decode.

pykern/util.py
```diff
@@ -14,6 +14,11 @@
 
 _DEFAULT_ROOT = "run"
 _DEV_ONLY_FILES = ("setup.py", "pyproject.toml")
+_STRANGE_CONTROL_CODES = [
+    bytes((n,))
```
robnagler, line 18: Didn't know this feature.

pykern/util.py
```diff
@@ -14,6 +14,11 @@
...
+    for n in range(0, 32)
+    if bytes((n,)) not in b"\x07\x08\t\n\x0b\x0c\r\x1b"
```
robnagler, line 20: instead of a byte string, simple compare the number to a set of the numbers. You could document them:

`_VALID_ASCII_CONTROL_CODES = frozenset(0x7, 0x8,,,)`

pykern/util.py
```diff
@@ -14,6 +14,11 @@
 
 _DEFAULT_ROOT = "run"
 _DEV_ONLY_FILES = ("setup.py", "pyproject.toml")
+_STRANGE_CONTROL_CODES = [
```
robnagler, line 17: `_INVALID_ASCII_CONTROL_CODES = frozenset(...`

pykern/util.py
```diff
@@ -65,3 +70,48 @@ def _check_dev_files(root):
...
+    Returns:
+        bool: True if bytes_data is likely pure text, false if likely binary
+    """
+
+    def _is_accepted_control_code_ratio(value):
+        c = 0
+        for i in range(len(value)):
+            if bytes([value[i]]) in _STRANGE_CONTROL_CODES:
+                c += 1
+        return (c / len(value)) < (1 / 3)
```
robnagler, line 91: `1/3` is a repeating decimal. It's probably clearer to just say `.33`.

pykern/util.py
```diff
@@ -65,3 +70,48 @@ def _check_dev_files(root):
...
+        for _ in range(4):
+            if _try(b):
+                return True
+            if len(b) <= 1:
+                return False
+            b = b[:-1]
+        return False
+
+    if _valid_unicode(value, test_size):
+        return _is_accepted_control_code_ratio(value)
```
robnagler, line 116: This has to be on the decoded version of the str. You'll have to ord() each utf8 character. You might have a check if the ord is above 127 or something.

tests/util_test.py
```diff
@@ -0,0 +1,27 @@
...
+        not util.is_pure_text(
+            b"\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f"
+        ),
+        not util.is_pure_text(b"\xd4\x16\xc0\xd6\xec\xbf\x92\xe6\x84T\xc9 \xe9\xbf"),
+        util.is_pure_text(b"This is example text"),
+        util.is_pure_text(b"\x07\x08\t\n\x0b\x0c\r\x0e\x0f"),
+        util.is_pure_text(a + "¡".encode("utf-8"), 2),
+        util.is_pure_text(a + b"\xf0\x9f\x8c\xae", 2),
+    ]:
+        pkunit.pkok(case, "is_pure_text test case failed")
```
robnagler, line 27: call is_pure_text here so you can print `value` in the pkok.

robnagler (changes requested):
Looks good. Minor style comments.

pykern/pkio.py
```diff
@@ -128,6 +129,21 @@ def has_file_extension(filename, to_check):
     return e in to_check
 
 
+def is_pure_text(filename, test_size=512):
+    """Uses pykern.util.is_pure_text to guess whether file is pure text
```
robnagler, line 133: Explain this without using "pure_text". You can say using is_pure_text, but then you might say: 

Read test_size bytes of path to determine if it is text file

See `pykern.util.is_pure_text` for the details on the heuristic used to test the first text_size bytes.

pykern/pkio.py
```diff
@@ -128,6 +129,21 @@ def has_file_extension(filename, to_check):
...
+
+    Args:
+        filename (str|py.path.local): file to check
```
robnagler, line 136: py.path is sufficient.

pykern/pkio.py
```diff
@@ -128,6 +129,21 @@ def has_file_extension(filename, to_check):
...
+        test_size (int): size of bytes chunk being tested
```
robnagler, line 137: language not right

pykern/pkio.py
```diff
@@ -128,6 +129,21 @@ def has_file_extension(filename, to_check):
     return e in to_check
 
 
+def is_pure_text(filename, test_size=512):
```
robnagler, line 132: path, not filename

pykern/util.py
```diff
@@ -65,3 +66,51 @@ def _check_dev_files(root):
         # Don't run from an install directory
         r = pkio.py_path()
     return pkio.mkdir_parent(r.join(_DEFAULT_ROOT))
+
+
+def is_pure_text(value, test_size=512):
+    """Guesses if value is pure text using heuristics.
```
robnagler, line 72: Describe the heuristics a bit.

pykern/util.py
```diff
@@ -65,3 +66,51 @@ def _check_dev_files(root):
...
+
+    Args:
+        value (bytes): bytes data
+        test_size (int): size of bytes chunk being tested
```
robnagler, line 76: I don't think `test_size` is relevant here. You have `value` and `len(value)` is what you are interested in. The algorithm should backup a few bytes to see if it can pass utf8 decoding.

The only problem you have here is if the value passed from pkio  is shorter than its test_size. That's a tricky relationship. I think what you really want is a flag "maybe_truncated" that pkio passes. If it is not truncated, then you don't backup, because you have the whole string here. The is_truncated flag is set in pkio by reading test_size + 1 and then checking if the read value is test_size + 1. If it isn't, then it isn't truncated. You only pass `value[:test_size]`, i.e. `util.is_pure_text(b[:test_size], is_truncated=len(b) > test_size)`.

pykern/util.py
```diff
@@ -65,3 +66,51 @@ def _check_dev_files(root):
...
+    """
+
+    def _is_accepted_control_code_ratio(text_value):
+        c = 0
+        for char in text_value:
+            if ord(char) == 0:
+                return False
+            if ord(char) < 32 and ord(char) not in _VALID_ASCII_CONTROL_CODES:
+                c += 1
+        return (c / len(text_value)) < 0.33
```
robnagler, line 89: Document 0.33 as ACCEPTABLE_CONTROL_CODE_RATIO

robnagler (changes requested):
I realize this is perfectionism, but designing good APIs is hard so I'd like to keep working on this.

pykern/pkio.py
gurhar1133: is "size of the bytes chunk read from filename" better?

gurhar1133: Or "size of the bytes chunk read from the start of filename"?

tests/util_test.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
```
robnagler, line 1: don't need coding

tests/util_test.py
```diff
@@ -0,0 +1,48 @@
...
+"""PyTest for :mod:`pykern.util`
+
+:copyright: Copyright (c) 2022 Bivio Software, Inc.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+import pytest
```
robnagler, line 7: pytest is an unnecessary import.

pykern/pkio.py
```diff
@@ -17,6 +17,7 @@
 import os
 import os.path
 import py
+import pykern.util
```
robnagler, line 20: remove `__future__` and coding above.

pykern/pkio.py
```diff
@@ -128,6 +129,23 @@ def has_file_extension(filename, to_check):
     return e in to_check
 
 
+def is_pure_text(filepath, test_size=512):
+    """Read test_size bytes of filepath to determine if it is likely a text file.
+    See pykern.util.is_pure_text for the details on the heuristics used to
```
robnagler, line 134: space line above and link with backticks and remove unnecessary words.

```
See `pykern.util.is_pure_text` for the heuristics used to
```

pykern/util.py
```diff
@@ -12,8 +12,10 @@
 import sys
 
 
+_ACCEPTABLE_CONTROL_CODE_RATIO = 0.33
```
robnagler, line 15: remove coding above.

pykern/util.py
```diff
@@ -65,3 +67,59 @@ def _check_dev_files(root):
...
+    Args:
+        value (bytes): bytes data
+        is_truncated (bool): whether or not value has been truncated
+
+    Returns:
+        bool: True if bytes_data is likely pure text, false if likely binary
+    """
+
+    def _is_accepted_control_code_ratio(text_value):
+        c = 0
```
robnagler, line 95: `c` could be `n` and `char` could be `c`. Having `char` doesn't really add much value.

pykern/util.py
```diff
@@ -65,3 +67,59 @@ def _check_dev_files(root):
...
+        for char in text_value:
+            if ord(char) == 0:
+                return False
+            if ord(char) < 32 and ord(char) not in _VALID_ASCII_CONTROL_CODES:
+                c += 1
+        return (c / len(text_value)) < _ACCEPTABLE_CONTROL_CODE_RATIO
+
+    def _try(chunk):
```
robnagler, line 103: `_try`  => `_try_utf8`.

pykern/util.py
```diff
@@ -65,3 +67,59 @@ def _check_dev_files(root):
...
+        try:
+            return chunk.decode("utf-8", "strict")
+        except UnicodeDecodeError:
+            return False
+
+    def _unicode_decoded(value):
```
robnagler, line 109: `_unicode_decoded` is `_ut8_decoded`. Technically Unicode is the general standard for utf8, utf16, etc.

tests/pkio_test.py
```diff
@@ -59,6 +59,15 @@ def test_has_file_extension():
     pkeq(False, pkio.has_file_extension("filename_with_no_extension", "json"))
 
 
+def test_is_pure_text():
```
robnagler, line 62: remove future and coding above.

tests/util_test.py
```diff
@@ -0,0 +1,48 @@
...
+from pykern.pkcollections import PKDict
```
robnagler, line 8: PKDict import in function. Avoid outer scope imports in tests.

tests/util_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def test_is_pure_text():
+    from pykern import util
+    from pykern import pkunit
+
+    a = "a".encode("utf-8")
+    for case in [
+        PKDict(
```
robnagler, line 17: There is a subtlety here with for loops in tests. When a for loop fails in a test, you don't get the case printed. In this case, that might be ok. However, there's a lot of code here anyway, because of the PKDict. If you have a lot of repeated code in a local context, there's almost no value in using named dicts. Someone reading it has to pass through the boiler plate. Rather, try to model on pkunit's expect on left and actual on right, and create a nested functions and default the common case (is_truncated=False) and don't encode unless you have to, e.g.

```py
def _false(value, is_truncated=False):
    punit.pkeq(False, util.is_pure_text(value, is_truncated=is_truncated))
def _true(value, is_truncated=False):
    punit.pkeq(True, util.is_pure_text(value, is_truncated=is_truncated))
_false(b"\0 one null causes failure in valid text")
_false(False, a + b"\xc2")
_false(False, bytes(range(1, 0x20))
etc.
```

What this will show you is that you are missing cases, e.g. is_truncated=True and expect=False. There may be others missing.

robnagler (changes requested):
Some cleanup and improving the test.

tests/util_test.py
```diff
@@ -0,0 +1,30 @@
...
+    def _true(value, is_truncated=False):
+        pkunit.pkeq(True, util.is_pure_text(value, is_truncated=is_truncated))
+
+    a = "a".encode("utf-8")
+    _false(b"\0 one null causes failure in valid text")
+    _false(a + b"\xc2")
+    _false(bytes(range(1, 0x20)))
+    _false(b"\xd4\x16\xc0\xd6\xec\xbf\x92\xe6\x84T\xc9 \xe9\xbf")
+    _false(a + b"\xc2\xc2\xc2\xc2", is_truncated=True)
+    _false(b"\x01" * 33 + b"\x07" * 67)
```
robnagler, line 24: Good boundary condition check. Maybe document the ratio change here?

robnagler (approved):
I'll let you merge. A bit of documentation on the cases would be good.

PR #435, opened 2024-01-18
Fix #434 support for radiasoft/sirepo#6636

### LINKED ISSUE #434: support for sirepo#6636
FileNotFoundError needs to be matched by pkio.exception_is_not_found.

robnagler:
- pkdexc accepts exc_info arg
- Add FileNotFound to exception_is_not_found
- add pkinspect.caller_func_name
- added pkunit.unbound_localhost_tcp_port and LOCALHOST_IP

### DIFF
--- a/pykern/pkunit.py
+++ b/pykern/pkunit.py
@@ -284,7 +284,7 @@ def file_eq(expect_path, *args, **kwargs):
 
 
 def unbound_localhost_tcp_port(start, stop):
-    """Looks for AF_INET SOCK_STREAM port for which bind fails
+    """Looks for AF_INET SOCK_STREAM port for which bind succeeds
 
     Args:
         start (int): first port
--- a/pykern/pkunit.py
+++ b/pykern/pkunit.py
@@ -298,12 +298,14 @@ def unbound_localhost_tcp_port(start, stop):
             s.bind((LOCALHOST_IP, int(port)))
         return port
 
-    for p in random.sample(range(start, stop), 10):
+    for p in random.sample(range(start, stop), 100):
         try:
             return _check_port(p)
         except Exception:
             pass
-    raise ValueError(f"unable find port in range={start}-{stop} ip={LOCALHOST_IP}")
+    raise ValueError(
+        f"unable find port random sample range={start}-{stop} tries=100 ip={LOCALHOST_IP}"
+    )
 
 
 def is_test_run():
--- a/pykern/pkunit.py
+++ b/pykern/pkunit.py
@@ -18,7 +18,9 @@ import json
 import os
 import py
 import pytest
+import random
 import re
+import socket
 import subprocess
 import sys
 import traceback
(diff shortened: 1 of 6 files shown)


e-carlin:
Can you fix the title? I don't think #434 is the correct issue.

pykern/pkunit.py
```diff
@@ -278,6 +283,29 @@ def file_eq(expect_path, *args, **kwargs):
     _FileEq(expect_path, *args, **kwargs)
 
 
+def unbound_localhost_tcp_port(start, stop):
+    """Looks for AF_INET SOCK_STREAM port for which bind fails
```
e-carlin, line 287: Shouldn't it be "...for which bind __succeeds__" or maybe better "...port that we are able to bind to"?

robnagler: If the bind fails, the port is "unbound" and therefore free to use as a server.

pykern/pkunit.py
```diff
@@ -278,6 +283,29 @@ def file_eq(expect_path, *args, **kwargs):
...
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.bind((LOCALHOST_IP, int(port)))
+        return port
+
+    for p in random.sample(range(start, stop), 10):
+        try:
+            return _check_port(p)
+        except Exception:
+            pass
+    raise ValueError(f"unable find port in range={start}-{stop} ip={LOCALHOST_IP}")
```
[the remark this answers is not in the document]
robnagler: ee50221369

pykern/pkunit.py
e-carlin: I guess I don't understand what "bind fails" means? Fails implies some sort of error to me. From the code below it looks to me that if something is already bound to the ip:port then bind "fails". 

```bash
~$ lsof -i -P | grep 12345
~$ python <<EOF
> import socket

with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
    s.bind(("127.0.0.1", 12345))
    print("after bind")
EOF

> > > > > after bind
~$ python -m http.server 12345 &
[1] 26826
~$ Serving HTTP on 0.0.0.0 port 12345 (http://0.0.0.0:12345/) ...

~$ lsof -i -P | grep 12345
python  26826 vagrant    3u  IPv4 1547964      0t0  TCP *:12345 (LISTEN)
~$ python <<EOF
> import socket

with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
    s.bind(("127.0.0.1", 12345))
    print("after bind")
EOF

> > > > > Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
OSError: [Errno 98] Address already in use
```

robnagler: b3b10bb646. boolean confusion, sorry

e-carlin (approved):

PR #436, opened 2024-01-25
Fix #429 readthedocs build

### LINKED ISSUE #429: Make readthedocs conform to current build process
Breaking this out of #386 since it's fairly simple and the other sub-tasks will start from that point

mkeilman:
Updates pykern to conform to the new [readthedocs](https://readthedocs.org/) process:
- adds `.readthdocs.yml` with required settings (and see below)
- adds `docs/requirements.txt` and excepting it from `docs/.gitignore`
- creates a ReadTheDocs command class that encapsulates the methods previously invoked directly by `setup()` and writes `docs/conf.py`

These are the commands that readthedocs runs to build documents:
```
git clone --depth 1 https://github.com/radiasoft/pykern.git .
git fetch origin --force --prune --prune-tags --depth 50 refs/heads/<branch>:refs/remotes/origin/<branch>
git checkout --force origin/<branch>
git clean -d -f -f
cat .readthedocs.yml
asdf global python 3.9.18
python -mvirtualenv $READTHEDOCS_VIRTUALENV_PATH
python -m pip install --upgrade --no-cache-dir pip setuptools
python -m pip install --upgrade --no-cache-dir sphinx readthedocs-sphinx-ext
python -m pip install --exists-action=w --no-cache-dir -r docs/requirements.txt
<commands in .readthedocs.yml/build/jobs/pre_build>
cat docs/conf.py
python -m sphinx -T -E -b html -d _build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
```
In our case the only thing in `docs/requirements.txt` is ".", meaning install pykern itself including its requirements as listed in `setup.py._requires()`, and the only pre-build command is `python setup.py readthedocs` to call the new command class.

Another note: we still need sphinx-apidoc to generate the .rst files. It's actually available on readthedocs but we'll do it in the command class so we have control over the packages.

YAN: `python setup.py tox` fails even in master. It's not related to this issue so I did not address it.

YAN: We may be able to ditch `_readthedocs_fixup()` but I'm not sure...


robnagler:
Superseded by #488

PR #440, opened 2024-02-02
FIx #400 Stdout feedback if test skipped

### LINKED ISSUE #400: In pktest, give feedback when tests are skipped
It can happen that environment differences can cause local test cases to succeed while failing with ci, due to the former being skipped. It is noted in the log file but not on stdout, so there is no cue to the user to investigate. Supply that info to the user, either while the tests are running or at the end.

Benbenbenin0:
Instead of nothing:
```
 pass
passed=1
```

Now prints for each skipped case:
```
tests/pkcli/sim_test.pytests/pkcli/sim_test.py::test_init_and_run SKIPPED (need to be able ...)
tests/pkcli/sim_test.pytests/pkcli/sim_test.py::test_skip SKIPPED (testing skipped ...)
tests/pkcli/sim_test.pytests/pkcli/sim_test.py::test_skip2 SKIPPED (unconditional skip)
 pass
passed=1
```

Tested with modified `tests/pkcli/sim_test.py`, catches skip() with/without reason and skipif().

@mkeilman wasn't sure between just a warning or actually printing cases, wondering what you think.


robnagler:
> tests/pkcli/sim_test.pytests/pkcli/sim_test.py::test_init_and_run SKIPPED (need to be able ...)

Maybe construct in a way that looks like other log entryies:

tests/pkcli/sim_test.pytests/pkcli/sim_test.py:555:test_init_and_run

pykern/pkcli/test.py
```diff
@@ -18,6 +18,7 @@
 
 SUITE_D = "tests"
 _COROUTINE_NEVER_AWAITED = re.compile("RuntimeWarning: coroutine .+ was never awaited")
+_TEST_SKIPPED = re.compile(".+ SKIPPED \(.+\)")
```
robnagler, line 21: If you'd like to match more securely, include line start/end, and don't forget `r`:

`re.compile(r"^.+\s+SKIPPED\s+\(.+\)$", flags=re.MULTILINE)`

Also be loose with spaces on parsing. If the output changes to two spaces or tab or something else, `\s` is more robust.

robnagler (changes requested):
Butting in with some ideas.

robnagler:
Also would be good to write a test.

Benbenbenin0:
New output:
```
tests/pkcli/sim_test.py pass
tests/pkcli/sim_test.py::test_init_and_run SKIPPED (need to be able ...)
tests/pkcli/sim_test.py::test_skip SKIPPED (unconditional skip)
tests/pkcli/sim_test.py::test_skipif SKIPPED (with reason)
passed=1
```

Thanks @robnagler for the notes, 2 questions:
1) Do you have a suggestion for testing these user messages? Anything inside the skipped test is, well, skipped and I'm not sure about capturing stdout in another test.
2) test1_test.py already has pytest imported, is that the right file for this test?

robnagler:
> Do you have a suggestion for testing these stdout messages? Anything inside the skipped test is, well, skipped and I'm unsure whether to try capturing output in another test.

Hard to prove a negative, but if you see a specific output from a test that is skipped, then it wasn't skipped. The test could be like:

```py
def test_skip():
    from pykern import pkunit

    pkunit.pkfail("test_skip not skipped")
```

Then you search for "test_skip not skipped" and fail if it is. You shouldn't see a failure either.

> test1_test.py already has pytest imported out of the test_test's, is that the right file for this test?

Create a separate test file (test_data/3_test.py) for this and write a separate case test_skip_output. You probably want a passing test in that file so you see passing and skipped.

Benbenbenin0:
@robnagler two more things:
> Then you search for "test_skip not skipped" and fail if it is. You shouldn't see a failure either.
- I assumed this meant search the test_log, am I getting the path correctly in case::test_skip_output()?
- pkfail doesn't seem to print it's msg in the test log until after all test complete, searching for "FAILED" instead.

robnagler:
You should be able to capsys. See how test_test does that.

Benbenbenin0:
@robnagler seems like capsys only captures from the output stream within a test case, meaning I can't see the pkfail output from `test_skip` in a following case.

I'm unsure where you want to do the checking, did you mean:
1) add a file `test_data/tests/3_test.py` containing a skipped case that pkfails with a unique msg
2) add another case to the same file that passes so `3_test.py` has passing and skipped
3) in `test_test.py` or somewhere else add a case `test_skip_output` that fails if capsys output contains the unique msg

robnagler:
[f29013f](https://github.com/radiasoft/pykern/pull/440/commits/f29013f4c1f65664c38242aac00aea4ad4371f30) fixes the test. Look at the diffs to understand.

Running tests of test infrastructure is very confusing. What I did first was break down test_simple to individual parts to see which part was throwing the exception (incorrectly). The test for exceptions being thrown during conformance clutters things so I like having one pkunit conditional at a time until I understand what's going on.

robnagler:
@Benbenbenin0 [7f8133a](https://github.com/radiasoft/pykern/pull/440/commits/7f8133a173ae85b0d56712132272f3a3b3c93dda) fixes an interesting problem. Like I said, testing test infrastructure is hard.

Benbenbenin0:
@robnagler I reviewed [f29013f](https://github.com/radiasoft/pykern/pull/440/commits/f29013f4c1f65664c38242aac00aea4ad4371f30), makes sense just two questions:
- Why test 1 & 3 then 2 & 3? Is this to test skips within a failing and passing test sequence?
- Why test 1 & 3 individually as well? These are all testing the basic function of pytest, is this to cover individual testing or something more redundant?

I also looked at [7f8133a](https://github.com/radiasoft/pykern/pull/440/commits/7f8133a173ae85b0d56712132272f3a3b3c93dda), makes sense why the CI was failing.

Still need a review from you to merge.

robnagler:
> Why test 1 & 3 then 2 & 3? Is this to test skips within a failing and passing test sequence?

Yes.

> Why test 1 & 3 individually as well? These are all testing the basic function of pytest, is this to cover individual testing or something more redundant?

We're making sure that calling default_command() with a single test (vs a directory) will produce different results. This definitely testing workflow inside pkcli.test.

robnagler (approved):

PR #442, opened 2024-02-05
Fix #441 do not touch `__deepcopy__` memo object

### LINKED ISSUE #441: `PKDict.__deepcopy__` treat memo as opaque

robnagler:


rorour (approved):
Looks good but I don't understand the purpose of `rv = self.copy()` rather than `rv = self.__class__()`. I'm not sure how you feel about dictionary comprehensions but we could even get rid of the temp var

```python3
    def __deepcopy__(self, memo):                                                                         |
        return PKDict(                                                                                    |
            {copy.deepcopy(k, memo): copy.deepcopy(v, memo) for (k, v) in self.items()}                   |
        )
```

rorour:
> Technically we shouldn't need deepcopy, but python does the [wrong thing](https://github.com/python/cpython/blob/main/Lib/copy.py#L217) with subclasses of dict:

Wow.

Makes sense, thank you for the explanation. I will add some documentation (#443)

PR #447, opened 2024-02-29
fix # 443: document PKDict.__deepcopy__()

rorour:


pykern/pkcollections.py
```diff
@@ -51,6 +51,8 @@ def __copy__(self):
         return self.__class__(self)
 
     def __deepcopy__(self, memo):
+        # use self.copy() instead of self.__class__(), because subclasses
+        # of PKDict may have constructors with different numbers of parameters
```
robnagler, line 55: I think this should be a docstring. It should explain its arguments even though they are "standard". Also explain the coupling with `__copy__` and document it. A bit more filler text in the description.

robnagler (changes requested):

pykern/pkcollections.py
rorour: 73403f857b02d89c940fc30f8c029f44e9621306

pykern/pkcollections.py
```diff
@@ -51,6 +51,18 @@ def __copy__(self):
         return self.__class__(self)
 
     def __deepcopy__(self, memo):
+        """
+        Override default deepcopy behavior for subclasses of dict. Relies on self.copy()
```
robnagler, line 55: Your explanation got me curious. It seems we don't need `__deepcopy__` any more. It may have been a bug in an older Python. I removed `__deepcopy__`, and all the tests in pykern and sirepo passed.  So instead of updating the docstring, let's delete `__deepcopy__`.

We should document `__copy__` and `copy`, because that's necessary, since `dict.copy` still always returns a `dict` even for subclasses so `__copy__` is necessary to allow subclasses to override that do not allow a constructor with no arguments.

robnagler (changes requested):
Sorry for the flip flopping...

pykern/pkcollections.py
rorour: self.__copy__ seems unnecessary to me now. I think PKDict only needs need to override self.copy, and subclasses with extra parameters can just override it too.

robnagler: True. We need to be careful about breakage so we need to change all `__copy__` calls to `copy` at the point of use before updating PKDict.

rorour: Code search doesn't return any uses of `__copy__` outside of the one in this module. Is it being used somewhere else?

robnagler: I did use it recently, then I deleted the code. I don't see anything either so go ahead and delete.

pykern/pkcollections.py
```diff
@@ -76,7 +67,13 @@ def __setattr__(self, name, value):
         super(PKDict, self).__setitem__(name, value)
 
     def copy(self):
-        return self.__copy__()
+        """Necessary because dict.copy will return a dict rather than PKDict.
```
robnagler, line 70: The first line is always [a summary](https://peps.python.org/pep-0257/#multi-line-docstrings) then you would give a more elaborate description. Something like:

```py
"""Override `dict.copy` to ensure the class of the return object is correct.

Necessary because dict.copy will return a dict rather than PKDict. Calls `self.__class__(self)` which makes a shallow copy. Subclasses do not accept this constructor form will need to override this method.

Returns:
     object: shallow copy of self
"""
```

We normally document `object` as the return value when it can be anything.

robnagler (changes requested):

pykern/pkcollections.py
rorour: 6089a9e08ba0a1b1ef9f12010460150ed5064269

robnagler (approved):

PR #461, opened 2024-03-26
Fix #444: fix warnings message

### LINKED ISSUE #444: Incorrect msg `FAILED due to warnings. See warnings summary`
This is confusing, because it doesn't show up in the warnings summary:

```txt
FAILED due to warning(s):
coroutine '_make_delegate_method.<locals>.method' was never awaited
```

gurhar1133:


pykern/pkcli/test.py
```diff
@@ -18,8 +18,9 @@
 
 SUITE_D = "tests"
 _COROUTINE_NEVER_AWAITED = re.compile("RuntimeWarning: coroutine .+ was never awaited")
+_FAILED_ON_WARNINGS = "\nFAILED due to:\n"
+_PASSED_LOG_PATTERNS = (r"=+ (\d+) passed.*=+", r"PASSED")
```
robnagler, line 22: these should be re.compile'd. I think you can leave PASSED, because you don't actually know which test passed. The test will output failed.

The first regex doesn't need parens around `\d+`. I don't understand the `=+`.

pykern/pkcli/test.py
```diff
@@ -188,8 +194,14 @@ def _try(output, restartable):
                     env=_env(restartable),
                 )
                 o = pkio.read_text(output)
-                if _COROUTINE_NEVER_AWAITED.search(o):
-                    pkio.write_text(output, o + _FAILED_ON_WARNINGS)
+                if m := _COROUTINE_NEVER_AWAITED.search(o):
```
robnagler, line 197: There may be multiple of these so you'll probably want to include that in the error message, e.g. 

```txt
Failed due to "coroutine was never awaited" here:
<text>
```

pykern/pkcli/test.py
```diff
@@ -18,8 +18,9 @@
 
 SUITE_D = "tests"
 _COROUTINE_NEVER_AWAITED = re.compile("RuntimeWarning: coroutine .+ was never awaited")
+_FAILED_ON_WARNINGS = "\nFAILED due to:\n"
```
robnagler, line 21: It's very specific so make the name specific. This doesn't need to be a constant.

pykern/pkcli/test.py
```diff
@@ -18,8 +18,9 @@
 
 SUITE_D = "tests"
 _COROUTINE_NEVER_AWAITED = re.compile("RuntimeWarning: coroutine .+ was never awaited")
```
robnagler, line 20: This should capture the entire line. You'll want to use "re.MULTILINE" flag on the compile.

pykern/pkcli/test.py
```diff
@@ -188,8 +194,14 @@ def _try(output, restartable):
                     env=_env(restartable),
                 )
                 o = pkio.read_text(output)
-                if _COROUTINE_NEVER_AWAITED.search(o):
-                    pkio.write_text(output, o + _FAILED_ON_WARNINGS)
+                if m := _COROUTINE_NEVER_AWAITED.search(o):
+                    pkio.write_text(
```
robnagler, line 198: I don't think you need to write the text. Just append to output.

tests/pkcli/test2_data/2.in/subprocess_test.py
```diff
@@ -0,0 +1,8 @@
+import sys
```
robnagler, line 1: There's already a test file. I'm a little confused why we need two tests that test the same thing.

robnagler (changes requested):

pykern/pkcli/test.py
gurhar1133: the "+=" was to match the "====== <number of passed> passed ====="

tests/pkcli/test2_data/2.in/subprocess_test.py
gurhar1133: Good point. I'll try to share

pykern/pkcli/test.py
gurhar1133: Don't I need to write to output since it is the log filename and if we find the runtimewarning in the log we have to write to that log file? Am I getting something mixed up?

robnagler: I see. You are rewriting the output to the log. I think we should just append to the log, not rewrite it. You could just add that failure and not remove the passed part of it.

pykern/pkcli/test.py
```diff
@@ -211,8 +212,18 @@ def _try(output, restartable):
             except Exception as e:
                 return _except(e, output, restartable)
             o = pkio.read_text(output)
-            if _COROUTINE_NEVER_AWAITED.search(o):
-                pkio.write_text(output, o + _FAILED_ON_WARNINGS)
+            if m := re.findall(_COROUTINE_NEVER_AWAITED, o):
+                pkio.write_text(
```
robnagler, line 216: The output is already written so append:

```py
with output.open(mode="a") as f:
    f.write("".join(f"ERROR: {x}\n" for x in m)
```

Extra newlines obscure other errors. At the moment, you are finding it hard to see this particular error. However, in your future, you'll be annoyed that this particular error is being highlighted, and the actual error is embedded somewhere else in the text.

There is no good solution to this particular problem. The types of errors we get are complex, especially because we are written concurrent code.

pykern/pkcli/test.py
```diff
@@ -17,9 +17,10 @@
 
 
 SUITE_D = "tests"
-_COROUTINE_NEVER_AWAITED = re.compile("RuntimeWarning: coroutine .+ was never awaited")
+_COROUTINE_NEVER_AWAITED = re.compile(
+    ".+\.py:\d+: RuntimeWarning: coroutine .+ was never awaited", flags=re.MULTILINE
```
robnagler, line 21: This regex can be simpler: `(.+ coroutine \S+ was never awaited.*)`. If this becomes too loose, we can always make it stricter. The failure mode of a too loose regular expression is going to be a failing test. That's an important concept.

tests/pkcli/test2_test.py
```diff
@@ -22,4 +21,6 @@ def test_cases(capsys):
         o, e = capsys.readouterr()
         pkio.write_text("stdout.txt", o)
         pkio.write_text("stderr.txt", e)
-        pkunit.pkre(expect[d.basename], o)
+        pkunit.pkre(
+            'FAILED due to these "coroutine was never awaited" warnings:\n1\).+2\)', o
```
robnagler, line 25: Too much text. Again, failure here is a failing test. Make the regex looser so that if we change the text, the test doesn't necessarily have to be updated. One thing that definitely should be checked is that on line 20, that the test actually failed. You expect this. 

It's still unclear to me that the subprocess is necessary.

robnagler (changes requested):

tests/pkcli/test2_test.py
gurhar1133: Are you thinking test_subprocess is not necessary if we arent using the warnings summary anymore and just appending the relevant info to the end of the log?

robnagler: Yea, I think you might be able to get away with just running it inline.

gurhar1133: Ok. Should I remove test_subprocess since if it is run inline it doesn't really do anything different from test_coroutine_never_awaited?

robnagler (approved):

Issue #462, opened 2024-03-27
Replace pkg_resources with importlib


robnagler:

robnagler:
This is really screwed up. They are adding things in 3.9 and deprecating in 3.12. We are going to have to change the model of pkresource to implement some form of caching of data, which aligns with how importlib.resources works. We'll have to provide backwards compatibility, because there's an assumption that the files can be in a wheel (zip) so you can't just symlink to those files.

I think at this time we can add a hook to pkcli to call clean up functions. I don't think pykern functions are invoked directly (e.g. uswgi) so they always start from a pkcli. We could check for this in side pkresource. 

pkresource will create a temporary directory (in /tmp is probably best, especially for docker) that caches resources and removes them on program exit (with pkcli hook).

Here's something that's really broken:

```sh
$ cd ~/src/radiasoft
$ python
Python 3.9.15 (main, Jan  5 2024, 23:53:42)
[GCC 12.2.1 20221121 (Red Hat 12.2.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib.resources as r
>>> for f in r.files("pykern").iterdir():
...     print(f)
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/resources.py", line 147, in files
    return _common.from_package(_get_package(package))
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/_common.py", line 14, in from_package
    return fallback_resources(package.__spec__)
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/_common.py", line 18, in fallback_resources
    package_directory = pathlib.Path(spec.origin).parent
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/pathlib.py", line 1082, in __new__
    self = cls._from_parts(args, init=False)
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/pathlib.py", line 707, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/pathlib.py", line 691, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I don't know how we'll work around this.

PR #463, opened 2024-03-27
Partial #462 remove pkg_resources from pksetup

robnagler:
- Added PYKERN_PKCLI_TEST_IGNORE_WARNINGS=1 to avoid deprecation warnings on tests. These are normally suppressed, but pytest enables all warnings.

### DIFF
--- a/pykern/pkcli/test.py
+++ b/pykern/pkcli/test.py
@@ -1,16 +1,15 @@
-# -*- coding: utf-8 -*-
 """run test files in separate processes
 
 :copyright: Copyright (c) 2019 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
+
 from pykern import pkconfig
 from pykern import pkio
 from pykern import pksubprocess
 from pykern import pkunit
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp
-import itertools
 import os
 import pykern.pkcli
 import re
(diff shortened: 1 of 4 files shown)


pykern/pkcli/test.py
```diff
@@ -171,31 +173,57 @@ def _except(exc, output, restartable):
...
         def _fail(output):
             self.failures.append(output)
             return f"FAIL {output}"
 
+        def _ignore_warnings():
+            if not _cfg.ignore_warnings:
+                return []
+            return list(
+                # This undoes what pytest does
+                itertools.chain.from_iterable(
```
gurhar1133, line 185: Would list.extend() be better than chain.from_iterable()?

Also, I remember running into problems trying to do ("-W", f"error::{runtime_warning_and_regex}"), but looking at how you pulled this off I think it might be worth another try.

gurhar1133 (approved):
LGTM. Approved, just left a comment on alternatives to chain.from_iterable, but not important.

Also, I remember running into problems trying to do ("-W", f"error::{runtime_warning_and_regex}"), but looking at how you pulled this off I think it might be worth another try.

pykern/pkcli/test.py
robnagler: chain.from_iterable flattens the list by one level. sum reduces a list. list.extend would require a temporary. It's debatable. I like functional that eliminates a temporary.

robnagler: It's better 272f257e85dbb3accef7f59de69d94fc1f7bc328.

PR #470, opened 2024-04-10
fix #469: add include_dirs option to pkio.walk_tree

### LINKED ISSUE #469: add include_dirs option to pkio.walk_tree
This would be useful for me in checking whether a user dir exists in sirepo tests

rorour:


robnagler:
Initially, walk_tree was recursive, and sorted_glob was not. walk_tree had a filter, sorted_glob did not. They were for two different purposes.

walk_tree takes a dirname and sorted_glob takes a glob. These are different approaches. walk_tree will return dot files, for example, and sorted glob won't.

sorted_glob is only recursive with `**`, which was added in 7081cb4a. key sorting was added in 4328a2a8. 

docstrings should refer to each other. The intentions were initially different. @e-carlin I don't remember why sorted_glob was expanded. Do you?

rorour:
@robnagler ready for review

pykern/pkio.py
```diff
@@ -285,14 +271,17 @@ def save_chdir(dirname, mkdir=False, is_pkunit_prefix=False):
...
 
-def sorted_glob(path, key=None):
-    """sorted list of py.path.Local objects
+def sorted_glob(pattern, key=None):
+    """Returns sorted list of files & dirs matching pattern.
 
-    Use '**' in path for a recursive search
+    Use '**' in pattern for recursive search, else, use * as wildcard.
+    Sorts using key if provided, else in ascending order.
+    Doesn't include hidden files unless . in pattern.
```
robnagler, line 279: unless dot "." is included explicitly at the start of a path component. 

BTW, If you do '.*' it only includes dot files. The same is true of '.**'. '**' does not include dot files even though it is recursive. Might be good to mention usage: sorted_glob is not usually recursive, and for recursive, it's probably better to use walk_tree and a file_re.

pykern/pkio.py
```diff
@@ -285,14 +271,17 @@ def save_chdir(dirname, mkdir=False, is_pkunit_prefix=False):
             pkunit_prefix = prev_ppp
 
 
-def sorted_glob(path, key=None):
-    """sorted list of py.path.Local objects
+def sorted_glob(pattern, key=None):
```
robnagler, line 274: Changing a parameter name is dangerous from a keyword arg perspective. I'll accept this, but I think it's important to note.

I was curious: glob.glob uses pathname, which is why it is just path here. C & Ruby use pattern. Go uses glob. Perl uses expr.

robnagler (changes requested):

pykern/pkio.py
robnagler: sorted_glob is fine for what it was intended for: pattern match some files. Adding the recurse=True was a mistake.

My impression is that `**` is expensive, but really depends on the implementation. It was [faster in 2012](https://stackoverflow.com/a/27565420) than os.walk, but then read the last comment about how 3.5 fixed the performance. 

These type of performance differences matter less than functionality. sorted_glob was, again, for a quick glob match that was sorted (Py glob doesn't sort), e.g. find all *.log files in order time order. There are uses of '**' mostly in tests. The one case in Sirepo is files_to_watch_for_reload probably would be better off with walk_tree, since the query is being executed NxM when it could be executed once with a combined regex for the paths to match. 

In the end, it's like a lot of things: there's more than one way, and we have personal preferences or what we can remember to pick one or the other. Programmers don't often agree on these things. :)

BTW, glob include_hidden was added in 3.11, which is false by default.

PR #475, opened 2024-04-26
Fix #474 fconf parser accepts base_vars

### LINKED ISSUE #474: fconf: base_vars
Parser takes base_vars which are an initial set of vars. For rsconf, this would be rsconf_db.db_d, etc.

robnagler:


pykern/fconf.py
```diff
@@ -205,7 +206,10 @@ def __init__(self, files):
             except Exception as e:
                 pykern.pkinspect.append_exception_reason(e, f"fconf.Parser.file={f}")
                 raise
-        e = _Evaluator(parser=self)
+        e = _Evaluator(
+            global_fvars=PKDict() if base_vars is None else copy.deepcopy(base_vars),
```
gurhar1133, line 210: A little confused about the `copy.deepcopy(base_vars)`. Why not just `base_vars`?

gurhar1133 (approved):
Looks good. One comment

pykern/fconf.py
robnagler: Need to make a copy, because base_vars because the result (global_fvars) of the parser. This means it is destructively modified, which clients of most APIs don't expect.

PR #500, opened 2024-08-23
Fix #499 pkcli gets commands from class Commands

### LINKED ISSUE #499: pkcli get commands from Commands class if there
Ignore superclass methods

robnagler:
- Instantiate the class first
- Ignore superclass methods

### DIFF
--- a/pykern/pkcli/__init__.py
+++ b/pykern/pkcli/__init__.py
@@ -28,11 +28,13 @@ import argh.assembling
 import argparse
 import importlib
 import inspect
+import itertools
 import os
 import os.path
 import pkgutil
 import re
 import sys
+import types
 
 # Avoid pykern imports so avoid dependency issues for pkconfig
 from pykern import pkconfig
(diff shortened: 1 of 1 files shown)


pykern/pkcli/__init__.py
```diff
@@ -187,20 +189,42 @@ def _argh_name_mapping_policy():
...
-    """Extracts all public functions from `cli`
+    """Extracts all public functions or methods from `cli`
 
     Args:
-        cli (module): where commands are executed from
+        cli (object): where commands are executed from
 
     Returns:
-        list of function: public functions sorted alphabetically
+        list: cmomands sorted alphabetically
```
e-carlin, line 198: typo

pykern/pkcli/__init__.py
```diff
@@ -187,20 +189,42 @@ def _argh_name_mapping_policy():
...
     """
-    res = []
-    for n, t in inspect.getmembers(cli):
-        if _is_command(t, cli):
-            res.append(t)
-    sorted(res, key=lambda f: f.__name__.lower())
-    return res
+
+    def _exclude():
```
e-carlin, line 201: I think a longer name would be helpful. Maybe `_methods_of_base_classes`. I could guess what __bases__ is but had to look it up. I also don't like names that know how a function will be used.

robnagler: [c2074bc](https://github.com/radiasoft/pykern/pull/500/commits/c2074bc5baefa1061edd6f114b61338a96786c2a)

PR #530, opened 2024-10-03
Fix #529: Don't keep reference to tornado.httpclient.AsyncHTTPClient

### LINKED ISSUE #529: tornado.httpclient.AsyncHTTPClient causing unbounded growth in number of open sockets
Creating [a reference to an instance of AsyncHTTPClient](https://github.com/radiasoft/pykern/blob/8894b7c9c9f3c1f09772dc09123404e4bb794f11/pykern/http.py#L151) is causing unbounded growth in the number of open sockets. Eventually we run out of file descriptors. I opened an issue in Tornado about this https://github.com/tornadoweb/tornado/issues/3424. As a stop gap I'm going to re-structure our HTTPClient code to remove storing the reference in self._client which should fix the issue.

e-carlin:
By keeping a reference to AsyncHTTPClient in self._client it causes sockets opened by the client to never be freed. Eventually, we run out of file descriptors. Writing the code this way w/o self._client prevents that from happening. This is a stop gap until we can fix the underlying bug.

### DIFF
--- a/pykern/http.py
+++ b/pykern/http.py
@@ -148,7 +148,6 @@ class HTTPClient:
                 _VERSION_HEADER: _VERSION_HEADER_VALUE,
             }
         )
-        self._tornado = tornado.httpclient.AsyncHTTPClient(force_instance=True)
         self._request_config = http_config.get("request_config") or PKDict()
 
     async def call_api(self, api_name, api_args):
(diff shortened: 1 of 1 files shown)


e-carlin:
@moellep please test with the slactwin code to confirm you see the socket growth issue resolved.

moellep (approved):
Works great - I didn't see any file leaks running lsof with this branch.

robnagler (approved):

PR #543, opened 2025-01-15
Fix #539 http: add client_id and authentication

### LINKED ISSUE #539: pykern.http change protocol to authenticate/version on first msg
Javascript Websocket object doesn't allow setting headers so switch to sending first message with auth info and version. This assumes msgpack is the protocol. Perhaps could assume JSON or just simple text: pykern-http <version> <token>.

robnagler:
- Removed backwards compatibility of implicit connect in call; Only one client, which is easily fixed, and connect is an important explicit step, which is async.


robnagler:
@e-carlin hold off. Need to add something.

robnagler:
@e-carlin ready for a review. The tests were failing due to the parallelization and pksubprocess race condition unrelated to this PR. I think I made it more reliable. #544 will be a better fix, but not working on that now, since it's really an unlikely situation.

pykern/http.py
```diff
@@ -406,7 +473,9 @@ def _api(call):
             return None
 
         async def _call(call, api, api_args):
-            with pykern.quest.start(api.api_class, self.server.attr_classes) as qcall:
+            with pykern.quest.start(
+                api.api_class, self.server.attr_classes, **pkdp(_quest_kwargs())
```
e-carlin, line 477: pkdp

e-carlin (approved):
One pkdp. Other than that LGTM

pykern/http.py
robnagler: 187ae8a

Issue #613, opened 2025-09-17
Should ActionLoop._dispatch_action be locked?


robnagler:
The lock is on object life cycle. If destroy is called during dispatch (e.g. and exception), the thread will deadlock (or get an exception?). Since the lock is only about the object life cycle, it doesn't really help the actions that are dispatched to be locked. The thread enforces only one action is called at a time and in order in the queue. Another option is that destroy is handled by the thread, which makes more sense from an ordering issue.

eloise-nebula:
If the lock is on the object's lifecycle, then its current implementation is overloaded. `_dispatch_action` and `destroy` cannot happen at the same time, meaning `ActionLoop` protects its own state. Presumably subclasses use `_destroy` in a way that could result in race conditions if not locked. The queue should be solely responsible for preventing race conditions by ordering, and the lock should be the same for managing object lifecycle.

In view of this, `_dispatch_action` should happen outside the lock, and the thread should handle `destroy`, but `destroy` should still immediately set the lifecycle flag `destroyed`. `ActionLoop` would only touch the the lock when dealing with `destroyed`. Calling `_destroy()` upon exiting the loop would be an easy way to implement this.

There's one caveat. The lifecycle flag `destroyed` is currently a public variable, meaning that any `ActionLoop` subclass or caller can see/change the lifecycle at any time with or without a lock. `ActionLoop` should restrict access to this flag if it's supposed to be solely responsible for its lifecycle. `destroyed` should be moved to `__destroyed`, and `ActionLoop` should provide the following for checking its lifecycle externally:
```
    def is_destroyed(self):
        with self.__lock:
            return self.__destroyed
```

Then, `destroy` would become:
```
    def destroy(self):
        try:
            with self.__lock:
                if self.__destroyed:
                    return
                self.__destroyed = True
```

robnagler:
The check for destroy can technically happen outside the lock, because it is getting a value and Python (even without the GIL) will either set a value or not. Writing destroy has to happen in the lock. Reading destroyed doesn't necessarily have to happen, since setting destroyed is not actually destroying the object and all actions (including destroy) are ordered inside the thread.

PR #614, opened 2025-09-17
Issue 613: Locking _dispatch_action

eloise-nebula:
A rough sketch of my solution to https://github.com/radiasoft/pykern/issues/613. Moves `_dispatch_action` and `_destroy` out of locks, but ensures that these two will never run simultaneously by letting the thread handle sequencing. Thread still exits on `self.__destroyed` such that no additional work can be done. If `_dispatch_action` is doing significant work it could take a while to exit, but that was already a problem. Plus, the new `is_destroyed` function provides a threadsafe way of checking the lifecycle flag from within `_dispatch_action`.

### DIFF
--- a/pykern/pkasyncio.py
+++ b/pykern/pkasyncio.py
@@ -49,7 +49,7 @@ class ActionLoop:
 
     def __init__(self):
         # All these attributes must exist even after destroy()
-        self.destroyed = False
+        self.__destroyed = False
         self.__lock = threading.Lock()
         self.__actions = queue.Queue()
         # You can join the thread to block another (e.g. main) thread from exiting
@@ -68,9 +68,7 @@ class ActionLoop:
         """Queue ``method`` to be called in loop thread.
 
         Actions are methods that (by convention) begin with
-        ``action_`` and are called sequentially inside `_start`. A
-        lock is used to prevent `destroy` being called during the
-        action and serializing activities within a single action.
+        ``action_`` and are called sequentially inside `_start`.
 
         Actions return ``None`` to continue on to the next
         action. `_LOOP_END` should be returned to terminate `_start`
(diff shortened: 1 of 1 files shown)


pykern/pkasyncio.py
```diff
@@ -174,25 +169,31 @@ def _start(self):
...
                 # Will be true if destroy called inside action (m)
-                if self.destroyed:
-                    return
+            with self.__lock:
+                if self.__destroyed:
+                    break
             # Action returned an external callback, which must occur outside lock
             if m:
                 self._dispatch_callback(m)
+            with self.__lock:
```
robnagler, line 192: I think this line should be outside the while loop. And, there is one case when the object is not destroyed but it exists, that's LOOP_END, which an action can return. It is used inside the upstream check, btw. If it is not destroyed, it should destroy itself, always.

pykern/pkasyncio.py
```diff
@@ -106,21 +104,18 @@ def action(self, method, arg):
...
-                    return
-                self.destroyed = True
-                self.__actions.put_nowait((None, None))
-                self._destroy()
-        except Exception as e:
-            pkdlog("error={} {} stack={}", e, self, pkdexc(simplify=True))
+        with self.__lock:
+            self.__destroyed = True
+
+    def is_destroyed(self):
```
robnagler, line 114: I would vote for not having this unless it's requested to be implemented. It's not clear that any client will actually need this. Rather the client should get an exception when it is destroyed and it tries to put an action, say.

pykern/pkasyncio.py
```diff
@@ -106,21 +104,18 @@ def action(self, method, arg):
...
-        try:
-            with self.__lock:
-                if self.destroyed:
-                    return
-                self.destroyed = True
-                self.__actions.put_nowait((None, None))
-                self._destroy()
-        except Exception as e:
-            pkdlog("error={} {} stack={}", e, self, pkdexc(simplify=True))
+        with self.__lock:
```
robnagler, line 111: `put_nowait((None, None))` has to happen after destroyed is set. This guarantees the thread will wake up after destroy has been called.

pykern/pkasyncio.py
```diff
@@ -174,25 +169,31 @@ def _start(self):
...
+                if self.__destroyed:
+                    break
             # Action returned an external callback, which must occur outside lock
             if m:
                 self._dispatch_callback(m)
+            with self.__lock:
+                # Can this happen?
+                if not self.__destroyed:
+                    return
+            self._destroy()
```
robnagler, line 196: Document this line since _destroy and destroy are so closely related.

robnagler (changes requested):
This looks great. Only a few comments.

pykern/pkasyncio.py
```diff
@@ -106,30 +104,26 @@ def action(self, method, arg):
...
-        try:
-            with self.__lock:
-                if self.destroyed:
-                    return
-                self.destroyed = True
-                self.__actions.put_nowait((None, None))
-                self._destroy()
-        except Exception as e:
-            pkdlog("error={} {} stack={}", e, self, pkdexc(simplify=True))
+        with self.__lock:
```
robnagler, line 117:
```py
if self.__destroyed:
    return
self.__destroyed = True
```

robnagler (changes requested):
Much better, thanks.

PR #639, opened 2026-03-12
fix #638 updates for SQLAlchemy 2

### LINKED ISSUE #638: upgrade to SQLAlchemy 2

moellep:
- wrap raw SQL in `text()` for SA 2 compatibility
- use `execution_options(autobegin=False)` for explicit transactions
- remove `<2` cap on SQLAlchemy dependency


moellep:
This needs more thought:
There should be a better way to access the row as a dictionary than going through ```._mapping```.
Clients of the API shouldn't be required to use ```._mapping```.
Also, sql_db methods which return a Row, like ```select_one()``` may want to return a PKDict directly.

Issue #648, opened 2026-05-14
sql_db.select compatibility with pkdict


eloise-nebula:
Transition to SQLAlchemy 2.0 broke the pattern:
```
pkdict(sql_db.select_one(foo))
```
which we use a lot in Slicops.

PR #649, opened 2026-05-14
Apply and test sql_db.select for PKDict compatability.

eloise-nebula:
SQLAlchemy Row objects do not unpack into dictionaries in 2.0 the way they did in 1.4. We often use that unpacking in slicops like this:

```
pkdict(sql_db.select_one(foo))
```

Migration guide suggests appending `.mappings()` to execute here:
https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#result-rows-act-like-named-tuples

See https://github.com/radiasoft/pykern/issues/648

### DIFF
--- a/tests/sql_db_test.py
+++ b/tests/sql_db_test.py
@@ -97,9 +97,10 @@ def _selects(meta):
 
     with meta.session() as s:
         r = s.select_one("t1", where=dict(a_name="Mildred"))
-        PKDict(r)
         pkunit.pkeq("hitch hiker's guide", r.a_text)
         t1, t2 = s.t.t1, s.t.t2
+        d = PKDict(r)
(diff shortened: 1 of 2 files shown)


tests/sql_db_test.py
```diff
@@ -92,10 +92,12 @@ def _meta(dir_path):
 
 def _selects(meta):
     from pykern import pkunit, pkdebug
+    from pykern.pkcollections import PKDict
     import sqlalchemy
 
     with meta.session() as s:
         r = s.select_one("t1", where=dict(a_name="Mildred"))
+        PKDict(r)
```
robnagler: This should test the keys that come back are what we expect (and not more).

robnagler (changes requested):
Thanks @eloise-nebula! Just a slight change.

robnagler (approved):

